PR #10, opened 2021-04-09
ADAPT-2183: Tailwind JIT.

sherakama:
# READY FOR REVIEW

# Summary
- Temporary fix until we can get some more time to flush out this approach. 

# Review By (Date)
- Whenever

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -9,9 +9,6 @@ import React from 'react';
 // Contexts.
 import { GlobalStateProvider } from './src/contexts/GlobalContext';
 
-// CSS
-import './src/styles/global.css';
-
 // Exports.
 export const wrapRootElement = ({ element }) => (
   <GlobalStateProvider>
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -9,6 +9,9 @@ import React from 'react';
 // Contexts.
 import { GlobalStateProvider } from './src/contexts/GlobalContext';
 
+// CSS
+import './src/styles/global.css';
+
 // Exports.
 export const wrapRootElement = ({ element }) => (
   <GlobalStateProvider>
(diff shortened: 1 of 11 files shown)


sherakama:
@yvonnetangsu 

The JIT compiler is definitely not 'seeing' any of the components from NPM packages on the production build. 

https://deploy-preview-10--stanford-alumni.netlify.app/test/yvonne/test-interior-page/

Nothing in the googles so far. Troubleshoot more next week.

yvonnetangsu:
> @yvonnetangsu
> 
> The JIT compiler is definitely not 'seeing' any of the components from NPM packages on the production build.
> 
> https://deploy-preview-10--stanford-alumni.netlify.app/test/yvonne/test-interior-page/
> 
> Nothing in the googles so far. Troubleshoot more next week.

Yeah, it's pretty weird and I haven't found anything online so far either. However, it does seem to "see" the local TW components, eg, the custom underline styles. So maybe we really could just selectively move over what we need from decanter v7 core since we have some issues with the plugins (not compatible with JIT) at the moment anyhow.

It doesn't see it on the netlify production build, but it works in the local production build (if I click on the html file in /public) which is weirder.

yvonnetangsu:
I'll make Cartman cross his cat fingers

sherakama:
Not too shabby: https://deploy-preview-10--stanford-alumni.netlify.app/test/yvonne/test-interior-page/

POC of scanning the build for the css render seems to work. There are issues with this approach and development but perhaps we can create a gatbsy plugin to do this.

sherakama:
On post-build event might be a way in. 
https://github.com/DSchau/gatsby-on-post-build-example/commit/8400ca5dda19f668ef8c04abfef76ab07ec6a3de

yvonnetangsu:
Ohhhhh awesome 🎉

sherakama:
@yvonnetangsu This should buy us some time until we can flush it out some more.

yvonnetangsu:
This is awesome - thank you for looking into this 😄

yvonnetangsu (approved):
Yay merging

PR #11, opened 2021-04-13
ADAPT-2114 ADAPT-2184: Hero, CTA Button, CTA Group components

yvonnetangsu:
# READY FOR REVIEW 

# Summary
- Hero, CTA Button and CTA group component

# Review By (Date)
- April 21 Wed

# Criticality

# Review Tasks

1. Look at https://deploy-preview-11--stanford-alumni.netlify.app/
2. Check each breakpoint (Note: there are some difference with Figma - some still waiting for final design specs - eg, button padding for mobile, some were discuss verbally). For this PR, just check that things look good generally for all breakpoints. Note: It may appears that for LG breakpoint and up, there is a lot of spacing at the top above the text, but that's because the local header will layer on the top of the hero banner.
3. Look at each of the button styles (I picked 3 different CTA button styles in the hero).
4. Play with Storyblok option on the Home page. Editor.js is being worked on right now - but you could preview by changing option, then hit save and refresh - the change will save and the preview error will go away.

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources


src/components/composite/hero.js
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <SbEditable content={props.blok}>
+      <Container className={`hero su-relative`} width='full'>
+        {props.blok.image.filename?.startsWith('http') && (
+          <FullWidthImage
+            filename={props.blok.image.filename}
+            imageFocus={props.blok.imageFocus}
+            className='hero su-absolute su-top-0 su-w-full su-h-full su-min-h-600 su-overflow-hidden'
+            imageClasses='su-w-full su-h-full su-object-cover'
```
sherakama, line 29: From a DX perspective, having two props for classes has me confused. I haven't looked at the FullWidthImage yet so I have my guesses but I would have suspected this component to be without a wrapper for the image element.

src/components/composite/hero.js
```diff
@@ -0,0 +1,70 @@
...
+            alt={props.blok.image.alt}
+          />
+        )}
+        <div className={`su-absolute su-block su-w-full su-h-full su-top-0 su-bg-gradient-to-b ${gradientFrom} su-to-saa-black`} aria-hidden='true' />
+        <Container className='su-relative su-rs-pt-9 su-rs-pb-4'>
+          <FlexBox direction='col' className='lg:su-mt-[190px]'>
+            <FlexCell className='su-text-center su-text-white'>
+              <p className='su-max-w-prose su-font-semibold su-leading-display su-text-m2 su-text-shadow-md md:su-text-m4 su-mx-auto su-mb-0'>{props.blok.sansSuper}</p>
+              <p className='su-max-w-prose su-font-serif su-leading-display su-text-m1 md:su-text-m2 su-text-shadow su-mx-auto su-mb-05em'>{props.blok.serifSuper}</p>
```
sherakama, line 38: Are any of these fields optional? If so, they should have checks in this template.

src/components/composite/hero.js
```diff
@@ -0,0 +1,70 @@
...
+            <FlexCell className='su-rs-mt-4'>
+              <CreateBloks blokSection={props.blok.cta} />
+            </FlexCell>
+            <FlexCell
+              grow={false}
+              className='su-text-center su-text-white su-font-semibold su-rs-mt-5 su-font-serif su-font-regular su-text-19 md:su-text-22'
+            >
+              <p className='su-mb-02em'>Scroll to explore</p>
+              <a href={`#${props.blok.scrollTo}`} className='su-block su-mx-auto su-w-fit su-group'>
+                <SrOnlyText srText='Jump to main content' />
```
sherakama, line 57: After reading this, I think I would rather see the format `<SrOnlyText>Jump to main content</SrOnlyText>` but no need to refactor.

sherakama: Is this ALWAYS going to jump to main content? Or should the text be configurable?

src/components/cta/ctaButton.js
```diff
@@ -0,0 +1,88 @@
...
+      <div className={`cta-button su-block su-mx-10 su-mb-20 last:su-mb-0 ${align}`}>
+        <SbLink
+          ref={ref}
+          link={props.blok.link}
+          attributes={props.blok.rel ? {rel: props.blok.rel} : {}}
+          classes={dcnb('su-inline-block su-w-fit su-group su-border-solid su-border-3 su-transition-colors su-no-underline su-underline-custom su-font-regular hover:su-underline focus:su-underline hover:su-shadow-md focus:su-shadow-md', ctaButtonStyle, ctaButtonSize)}
+        >
+          {props.blok.linkText}
+          {props.blok.srText &&
+          <span className='su-sr-only'>{` ${props.blok.srText}`}</span>
```
sherakama, line 72: Why not use the `SrOnlyText` component here?

src/tailwind/plugins/components/gradient-border/gradient-border.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Custom gradient border styles
+ */
+module.exports = function () {
+  return function ({ addComponents }) {
+    const components = {
+      '.gradient-border': {
+        borderImageSlice: '1',
+        borderImageSource: 'linear-gradient(to right top, #017E7C, #505EEC)'
+      },
```
sherakama, line 10: Any chance you are going to have more than one gradient border? Might want to think of a different name or to create a combination of all the classes using the predefined color options.

src/tailwind/plugins/components/shadow/text-shadow.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Text shadow styles
+ */
+module.exports = function () {
```
sherakama, line 4: This one would be a good candidate for the core Decanter product. Consider moving it into v7.

src/utilities/dataSource.js
```diff
@@ -6,9 +6,15 @@ export const bgPositionVertical = {
   bottom: 'su-bg-bottom',
 };
 
+export const ctaGroupDisplay = {
```
sherakama, line 9: Where did we leave off with pulling these with a static query?

src/components/media/fullWidthImage.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import transformImage from '../../utilities/transformImage';
+import getImageWidth from '../../utilities/getImageWidth';
+import { objectPosition } from '../../utilities/dataSource';
+import { dcnb } from 'cnbuilder';
+
+const FullWidthImage = ({ filename, element, imageClasses, alt, imageFocus, ...props}) => {
+  const Element = element ?? 'figure';
```
sherakama, line 8: Would we ever want the option for no wrapping element?

src/components/media/fullWidthImage.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <Element {...props}>
+      <img
+        {...(imgSrcset ? {srcSet: imgSrcset} : {})}
+        {...(imgSizes ? {sizes: imgSizes} : {})}
+        src={imgSrc}
+        className={dcnb(imageClasses, imgFocus)}
+        alt={alt}
+      />
```
sherakama, line 63: What about a props for the image as well for things like title, data, or aria attributess?

sherakama (changes requested):
Looks good. Overall, pretty minor feedback.

One visual thing I did notice was that the scroll down arrow was below the fold for me. Not sure if that was intentional or not.

src/components/composite/hero.js
yvonnetangsu: Yeah I know what you mean, though having a figure wrapper allows for object-fit (for image focus) and aspect ratio stuff - which is used a lot in components. Perhaps I could rename it 🤔

src/components/composite/hero.js
yvonnetangsu: Good catch - how did I forget this one geez

src/components/composite/hero.js
yvonnetangsu: I think it is always going to jump to main content since they're only going to use it on the homepage. Maybe leave it like this for now and if there's a need later, I'll make it configurable.

src/components/cta/ctaButton.js
yvonnetangsu: Yeah good point

src/tailwind/plugins/components/gradient-border/gradient-border.js
yvonnetangsu: I believe there's only going to be one gradient border for this SAA theme. I agree that if there will be different color options, then the classname should be different.

src/tailwind/plugins/components/shadow/text-shadow.js
yvonnetangsu: Good idea. Maybe we can get Jacob to help out 😄

src/utilities/dataSource.js
yvonnetangsu: For now, I like having them all in the one dataSource.js file since I can easily tweak them there (and easier to see), and works well with the JIT mode. I don't have any classnames on the storyblok side.

src/components/media/fullWidthImage.js
yvonnetangsu: I could split it into 2 components - the FullWidthImage (just the img), and the optional Figure component that acts as wrapper 🤔 . Maybe next sprint 😛

src/components/media/fullWidthImage.js
yvonnetangsu: Good point - I'll tackle this in the above mentioned splitting FullWidthImage ticket (will add).

yvonnetangsu:
> Looks good. Overall, pretty minor feedback.
> 
> One visual thing I did notice was that the scroll down arrow was below the fold for me. Not sure if that was intentional or not.

Yeah, it's a pretty large hero banner. Kerri did say that she doesn't intend for them to add every field to the banner, so it might become shorter. I suspect we will finetune spacing and things after adding in the masthead though 😃

src/components/cta/ctaButton.js
yvonnetangsu: Done, also used it in ctaLink

src/components/composite/hero.js
yvonnetangsu: All done

yvonnetangsu:
@sherakama Ready for another look. I'll work on the image component stuff next sprint. Thanks! Heads up I might be in and out today. Starting to feel weird chills.

yvonnetangsu:
@sherakama , I did some experiments and found that basically, the JIT modes doesn't like our custom link utilities when they have the `.hover:` typed prefixes. Those TW variant prefixes confuse the JIT mode and thus, in the global footer, when we add a class like `.hover:su-link-no-underline` to the `nav` element, it does not respect the custom utility that we add.

We want this:
```
.hover:su-link-underline {
  a:hover {
  text-decoration: underline;
  }
}
```

Instead, this is what it will generate:
```
.hover:su-link-underline {
  :hover a {
  text-decoration: underline;
  }
}
```

I tried a number of things, including moving them to "components" instead of "utilities", disable variants, manually write out everything and none of those methods really worked. So now I'm going with my last resort, which is to create a SAA `base.js` file, which adds to the Decanter `base.js` file. It includes overrides to some of the styles that we can't change (eg, we don't have access to the link classes in the global footer `nav` so can't use dedupe). Currently, the global footer is working using this method. Let me know what you think. I know it's not great, but at least the global footer is working 😬

src/utilities/dataSource.js
sherakama: > I don't have any classnames on the storyblok side

Ah right. PurgeCSS is going to be the bain of my existence for some time I foresee. We've got to give this a hard think as it isn't a great DX experience to have EVERY CSS STRING IN THE SOURCE CODE and can cause brittleness between our CMS and code.

yvonnetangsu: Argghh....yeah. It's definitely a bit of a pain with PurgeCSS and JIT.

yvonnetangsu: I did see another option that we haven't explored. It looks like for Storyblok option fields, instead of datasource, you could pick "eternal json" - maybe we can give that a try.

sherakama: Oh, that sounds like an interesting option. I'll create a ticket to explore that later.

sherakama: Ah, but that gets us roughly to the same place as the datasource as I want to have options configured on the storyblok side and not the code side. I suppose this might be a limitation we have to live with.

yvonnetangsu: Yeah - it's still roughly the same. I find the datasource stuff works pretty well most of the time since they tend to include classes that we'll need to write out explicitly anyway. Maybe we'll just need to generate our own safelist.txt file like TW suggested for the ones that we don't want to write out, like the grid column classes. Not ideal, but looks like TW is aware of the limitation that JIT mode doesn't allow safelisting classes, so they might add a feature later on to enable that.

src/components/composite/hero.js
sherakama: Break it out into two components? I'll let you deliberate on that and revise/refactor later.

src/components/composite/hero.js
sherakama: SG2M. One of the practices of XP (Extreme Programming) is to never write code for the future. That pairs well with Agile.

sherakama (approved):
GTG

PR #21, opened 2021-04-26
ADAPT-2116 ADAPT-2387 Section, Grid components

yvonnetangsu:
# READY FOR REVIEW
- (Edit the above to reflect status)

# Summary
- Add Grid and Section component for layouts
- This branch has pulled in Shea's Decanter-React refactor branch for safe purgeCSS
- This branch was originally cloned from my basic card branch

# Review By (Date)
- Probably better to merge the Basic Card branch first before reviewing this one

# Criticality
- How critical is this PR on a 1-10 scale? Also see [Severity Assessment](https://stanfordits.atlassian.net/browse/D8CORE-1705).
- E.g., it affects one site, or every site and product?

# Review Tasks

## Setup tasks and/or behavior to test

1. go to https://deploy-preview-21--stanford-alumni.netlify.app/test/yvonne/test-section/
2. Look at the different section/grid options
3. Pull down locally, nvm use, npm run develop, go to http://app.storyblok.com/#!/me/spaces/103430/stories/0/0/47236458
4. Look at and play with Storyblok options for Section and Grid

## Front End Validation
- [ ] Is the markup using the appropriate semantic tags and passes HTML validation?
- [ ] Cross-browser testing has been performed?
- [ ] Automated accessibility scans performed?
- [ ] Manual accessibility tests performed?
- [ ] Design is approved by @ user?

### Code
- [ ] Are the naming conventions following our standards?
- [ ] Does the code have sufficient inline comments?
- [ ] Is there anything in this code that would be hidden or hard to discover through the UI?
- [ ] Are there any [code smells](https://blog.codinghorror.com/code-smells/)?
- [ ] Are tests provided? eg (unit, behat, or codeception)

### Code security
- [ ] Are all [forms properly sanitized](https://www.drupal.org/docs/8/security/drupal-8-sanitizing-output)?
- [ ] Any obvious [security flaws or new areas for attack](https://www.drupal.org/docs/8/security)?

## General
- [ ] Is there anything included in this PR that is not related to the problem it is trying to solve?
- [ ] Is the approach to the problem appropriate?

# Affected Projects or Products
- Does this PR impact any particular projects, products, or modules?

# Associated Issues and/or People
- JIRA ticket(s)
- Other PRs
- Any other contextual information that might be helpful (e.g., description of a bug that this PR fixes, new functionality that it adds, etc.)
- Anyone who should be notified? (`@mention` them here)

# Resources
- [AMP Tool](https://stanford.levelaccess.net/index.php)
[...]

PR #22, opened 2021-04-29
ADAPT-2116: Suggested Edits.

sherakama:
# READY FOR REVIEW

Suggestions and edits for #21

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@heroicons/react": "^1.0.0",
     "@tailwindcss/aspect-ratio": "^0.2.0",
-    "decanter-react": "github:SU-SWS/decanter-react#ADAPT-2296",
+    "decanter-react": "github:SU-SWS/decanter-react#main",
     "gatsby": "3.2.1",
     "gatsby-plugin-anchor-links": "^1.2.0",
     "gatsby-plugin-manifest": "3.2.0",
@@ -38,13 +38,12 @@
   ],
   "license": "MIT",
   "scripts": {
-    "build": "gatsby build && npm run wind",
+    "build": "gatsby build",
     "develop": "gatsby develop",
     "start": "npm run develop",
     "serve": "gatsby serve",
     "clean": "gatsby clean",
-    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1",
-    "wind": "tailwindcss build ./src/styles/global.css -o ./public/styles/global.css"
+    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1"
   },
   "repository": {
     "type": "git",
--- a/src/components/cards/basicCard.js
+++ b/src/components/cards/basicCard.js
@@ -7,8 +7,26 @@ import { FlexBox, Heading } from 'decanter-react';
 import { dcnb } from 'cnbuilder';
 import { borderColors } from '../../utilities/dataSource';
 
-const BasicCard = (props) => {
-  const numCta = getNumBloks(props.blok.cta);
+const BasicCard = ({
+  blok: {
+    cta,
+    borderColor,
+    isRound,
+    isMinimal,
+    isLightText,
+    isBigHeadline,
+    align,
+    image: {
+      filename
+    },
+    imageFocus,
+    headline,
+    headingLevel,
+    text
+  },
+  blok
+}) => {
+  const numCta = getNumBloks(cta);
 
   let wrapperClasses = 'su-bg-white su-text-black su-border su-border-solid su-border-transparent-black su-shadow'
   let imageClasses;
@@ -17,9 +35,9 @@ const BasicCard = (props) => {
   let imageWrapperClasses = 'su-aspect-w-4 su-aspect-h-3';
 
   // Option to display image as round thumbnail with colored border
-  const borderColor = borderColors[props.blok.borderColor] ?? borderColors['digital-red'];
+  borderColor = borderColors[borderColor] ?? borderColors['digital-red'];
 
-  if (props.blok.isRound) {
+  if (isRound) {
     wrapperClasses = dcnb('su-rs-pt-3', wrapperClasses);
     imageWrapperClasses = dcnb('su-w-[14rem] su-h-[14rem] su-rs-ml-2 su-rounded-full su-border-[7px] su-border-solid su-overflow-hidden', borderColor);
     imageClasses = 'su-w-full su-h-full';
@@ -28,21 +46,21 @@ const BasicCard = (props) => {
   // Option to use "minimal" card variant
   let bodyPadding = 'su-rs-px-2 su-rs-pt-2 su-rs-pb-4'
 
-  if (props.blok.isMinimal) {
+  if (isMinimal) {
     wrapperClasses = '';
     bodyPadding = 'su-rs-pt-2';
     imageWrapperClasses = dcnb(imageWrapperClasses, {'su-rs-ml-2': false});
   }
 
   // Option to use light text
-  if (props.blok.isLightText) {
+  if (isLightText) {
     wrapperClasses = 'su-bg-transparent su-text-white';
   }
 
   // Option to make headline font larger
   let headlineSize = 'su-type-2';
 
-  if (props.blok.isBigHeadline) {
+  if (isBigHeadline) {
     headlineSize = 'su-type-3';
   }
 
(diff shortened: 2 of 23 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@heroicons/react": "^1.0.0",
     "@tailwindcss/aspect-ratio": "^0.2.0",
-    "decanter-react": "github:SU-SWS/decanter-react#ADAPT-2296",
+    "decanter-react": "github:SU-SWS/decanter-react#main",
```
sherakama, line 12: Merged this PR.

package.json
```diff
@@ -38,13 +38,12 @@
...
   "scripts": {
-    "build": "gatsby build && npm run wind",
+    "build": "gatsby build",
     "develop": "gatsby develop",
     "start": "npm run develop",
     "serve": "gatsby serve",
     "clean": "gatsby clean",
-    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1",
-    "wind": "tailwindcss build ./src/styles/global.css -o ./public/styles/global.css"
+    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1"
```
sherakama, line 46: Shouldn't need this now that we have purge safe markup.

package.json
```diff
@@ -38,13 +38,12 @@
...
   "license": "MIT",
   "scripts": {
-    "build": "gatsby build && npm run wind",
+    "build": "gatsby build",
     "develop": "gatsby develop",
     "start": "npm run develop",
     "serve": "gatsby serve",
     "clean": "gatsby clean",
-    "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1",
-    "wind": "tailwindcss build ./src/styles/global.css -o ./public/styles/global.css"
```
sherakama, line 47: Shouldn't need this now that we have purge safe markup.

src/components/composite/hero.js
```diff
@@ -6,71 +6,88 @@ import CreateBloks from '../../utilities/createBloks';
...
                            font='serif'
                            weight='bold'
                            className={`su-leading-tight su-tracking-normal su-text-shadow-lg su-mb-02em ${headlineSize}`}
                   >
-                    {props.blok.headline}
+                    {headline}
                   </Heading>
                 }
-                {props.blok.sansSub &&
-                  <p className='su-max-w-prose su-mx-auto su-text-20 md:su-text-m1 su-leading-display su-text-shadow su-mx-auto su-mb-0'>{props.blok.sansSub}</p>
```
sherakama, line 59: su-mx-auto is duplicated.

src/components/composite/hero.js
```diff
@@ -6,71 +6,88 @@ import CreateBloks from '../../utilities/createBloks';
...
             {numCta > 0 &&
-              <FlexCell className={props.blok.sansSub ? 'su-rs-mt-4' : ''}>
-                <CreateBloks blokSection={props.blok.cta}/>
+              <FlexCell className={sansSub ? 'su-rs-mt-4' : ''}>
+                <CreateBloks blokSection={cta}/>
               </FlexCell>
             }
             <FlexCell
               grow={false}
-              className='su-text-center su-text-white su-font-semibold su-rs-mt-5 su-font-serif su-font-regular su-text-19 md:su-text-22'
```
sherakama, line 70: su-font-semibold was overridden by su-font-regular.

src/components/composite/hero.js
```diff
@@ -6,71 +6,88 @@ import CreateBloks from '../../utilities/createBloks';
...
+              className='su-text-center su-text-white su-rs-mt-5 su-font-serif su-font-regular su-text-19 md:su-text-22'
             >
               <p className='su-mb-02em'>Scroll to explore</p>
-              <a href={`#${props.blok.scrollTo}`} className='su-block su-mx-auto su-w-fit su-group'>
```
sherakama, line 73: unnecessary interpolation

package.json
yvonnetangsu: Cool. Thanks!

src/components/composite/hero.js
yvonnetangsu: This one I think we might need to keep the original because we need the #

src/components/cards/basicCard.js
```diff
@@ -7,8 +7,26 @@ import { FlexBox, Heading } from 'decanter-react';
...
+    image: {
+      filename
+    },
+    imageFocus,
+    headline,
+    headingLevel,
+    text
+  },
+  blok
+}) => {
```
yvonnetangsu, line 28: Interesting....sure, I like it, making it clearer. Thanks for the refactor!

src/components/cards/basicCard.js
```diff
@@ -17,9 +35,9 @@ const BasicCard = (props) => {
   let imageWrapperClasses = 'su-aspect-w-4 su-aspect-h-3';
 
   // Option to display image as round thumbnail with colored border
-  const borderColor = borderColors[props.blok.borderColor] ?? borderColors['digital-red'];
+  borderColor = borderColors[borderColor] ?? borderColors['digital-red'];
```
yvonnetangsu, line 38: So it's ok to do this? ie, the value on the left side of the "=" having the same name as the prop (the prop.blok.borderColor)? Is it only ok now because borderColor is destructured? When I did that before, it would complain saying the variable name can't be the same as the prop name.

src/components/cta/ctaButton.js
```diff
@@ -6,33 +6,47 @@ import { buttonSize, buttonStyle, heroicon, textAlign } from '../../utilities/da
...
+  },
+  ref
+) => {
   // Button size
-  const ctaButtonSize = buttonSize[props.blok.size] ?? buttonSize['default'];
+  const ctaButtonSize = buttonSize[size] ?? buttonSize['default'];
 
   // Button style
-  const ctaButtonStyle = buttonStyle[props.blok.buttonStyle] ?? buttonSize['primary'];
+  const ctaButtonStyle = buttonStyle[propsButtonStyle] ?? buttonSize['primary'];
```
yvonnetangsu, line 28: Is this one handled different because "style" is a special word? I remember having issues before when naming a prop "style".

src/components/cards/basicCard.js
yvonnetangsu: I saw in below code that for example this one is handled this way:
`const myDisplay = ctaGroupDisplay[display] ?? ctaGroupDisplay['adjacent'];`

Is there a reason this one is different?

yvonnetangsu:
Thank you for the refactor, @sherakama ! Looks great - just those question for you to make sure I'm understanding it correctly.

src/components/composite/hero.js
sherakama: Ah, duh. 

`<a href={`#${scrollTo}`} ...`

src/components/cards/basicCard.js
sherakama: Ideally, the destructured prop and any other variables would have different names. I didn't get any complaints from the compiler when I re-set the prop variable with a new value but we really shouldn't be doing that. Best to either alias the destructured prop or to change the variable name in the code below. I can let you pick if you'd like to change it.

src/components/cta/ctaButton.js
sherakama: `style` is a reserved word. This was aliased due to the fact there was another thing called `buttonStyle`

src/components/cards/basicCard.js
yvonnetangsu: Oh that's great, I kind of prefer having a different variable name myself - it's a bit clearer for me.

src/components/cta/ctaButton.js
yvonnetangsu: Ohh right.... the datasource is also called buttonStyle 😂

yvonnetangsu:
I pulled down this branch locally and did `npm run develop`, but got a runtime error for going to an interior page in Storyblok for the title in SEO component, so instead of renaming the title in the prop to theTitle, I tried renaming the title in UseMetadata to siteTitle to see if it makes a difference, but I ended up with the same error just in a different place. It doesn't look to me that we did anything wrong so I'm a bit puzzled by this

Screen Shot 2021-04-28 at 11 33 05 PM

src/components/partials/seo.js
```diff
@@ -11,28 +11,28 @@ import transformImage from "../../utilities/transformImage";
...
+      <Helmet titleTemplate={`%s | ${title}`} title={theTitle}>
         <meta name='description' content={description} />
       </Helmet>
     )
   }
 
   // Use the title in SEO component, otherwise use the page title
-  let seoTitle = props.blok.seo.title || props.blok.title || '';
-  let ogTitle = props.blok.seo.og_title || seoTitle;
+  let seoTitle = seo.title || title || '';
```
yvonnetangsu, line 27:
```suggestion
  let seoTitle = seo.title || theTitle || '';
```
Small fixup

yvonnetangsu:
It also complains about the localFooter, but when I looked at the json of the interiorPage in Storyblok, localFooter is there
Screen Shot 2021-04-29 at 12 00 21 AM

For trouble shooting, I tried reverting the changes in the layout.js file (it has both the SEO and local footer), but that didn't make a difference for me 🤔

sherakama:
@yvonnetangsu see [`c0aa613` (#22)](https://github.com/SU-SWS/saa_alumni/pull/22/commits/c0aa61310ad46cc458c1f6106a1c6d046d034b03)

yvonnetangsu:
> @yvonnetangsu see [`c0aa613` (#22)](https://github.com/SU-SWS/saa_alumni/pull/22/commits/c0aa61310ad46cc458c1f6106a1c6d046d034b03)

Huh.... is interior page a special snowflake because we need to access the props of interior page from other components?

sherakama:
Layout component needs what it needs. It was looking for blok and previously when we destructured we pulled it out into its own variable. `blok` was not included in `rest` at that point. Passing down everything seemed appropriate here.

yvonnetangsu:
Thanks - merging this 🎈

PR #46, opened 2021-06-21
ADAPT-2735 Build Simple Image component

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Build Simple Image component

# Review By (Date)
- tbd

# Review Tasks

1. Check out this branch
2. Run local dev and review the Simple Image component at `/test/rebecca/simple-image-test ` or [Netlify preview](https://deploy-preview-46--stanford-alumni.netlify.app/test/rebecca/simple-image-test)

## Front End Validation

### Code

### Code security

## General

- [ADAPT-2735](https://stanfordits.atlassian.net/browse/ADAPT-2735)


src/components/media/simpleImage.js
```diff
@@ -0,0 +1,25 @@
...
+import SbEditable from "storyblok-react";
+import WidthBox from "../layout/widthBox";
+import RichTextRenderer from "../../utilities/richTextRenderer";
+
+const SimpleImage = ({
+  blok: { image: { filename } = {}, imageWidth, caption, captionAlign },
+  blok,
+}) => (
+  <SbEditable content={blok}>
+    <WidthBox width={imageWidth}>
```
sherakama, line 11: This looks and functions well, I just wonder about the HTML element and changing it to figure so we can use figcaption with the caption. @yvonnetangsu thoughts, and perhaps a quick SODA request?

yvonnetangsu: Yeah, I think it would be nice to use the figure element. The WidthBox uses the Grid component from Decanter React, but I'm totally fine with adding a `<figure>` wrapper around the image and `figcaption` and nest the `figure` within the WidthBox (so WidthBox can stay as is).

src/components/media/simpleImage.js
```diff
@@ -0,0 +1,25 @@
...
+      <img src={filename} alt="" />
+      {caption && (
+        <RichTextRenderer
+          wysiwyg={caption}
+          className={`su-media__caption su-text-${captionAlign} ${
+            imageWidth === "edge-to-edge" ? "centered-container" : ""
+          }`}
```
yvonnetangsu, line 18: You might not need all of these classnames here - can probably remove the su-media__caption (that's the BEM naming we use for Giving only) and the imageWidth class is probably not needed as well since the WidthBox should take care of the bounding box (but please double check because I could be wrong 😄 )

yvonnetangsu: Please wrap this whole RichTextRenderer wtihin a `<figcaption>`. Thanks!

yvonnetangsu: Please also add the `su-caption` classname to the figcaption - I think that would take care of the caption styling (it's from Decanter v7 under TW `components/typography/styles`)

yvonnetangsu:
@rebeccahongsf Looking great! A few comments added, and we'll probably groom this ticket a bit more after (eg, we might want to add the spacing top/bottom option in the component as well. Thanks! 😄

src/components/media/simpleImage.js
```diff
@@ -0,0 +1,27 @@
...
+import RichTextRenderer from "../../utilities/richTextRenderer";
+
+const SimpleImage = ({
+  blok: { image: { filename } = {}, imageWidth, caption, captionAlign },
+  blok,
+}) => (
+  <SbEditable content={blok}>
+    <WidthBox width={imageWidth}>
+      <figure>
+        <img src={filename} alt="" />
```
yvonnetangsu, line 13: I think we're going to need to pass an option alt as prop here. If the alt is provided, then it should show up, if not, if should render the alt="" tag 😄

yvonnetangsu (changes requested):
Thank you for the changes - works great 😄 . We'll likely groom this ticket some more tomorrow, but in the meantime, I left the comment about adding an alt as prop, and if you could please add some image resizing that'd be great. 

Please see the Giving example (first get the image width of the uploaded original image, then resize according to the width option they choose, and only resize if the original size is wider than the width option)
https://github.com/SU-SWS/ood_giving_site/blob/main/src/components/media/storyImage.js

We have the utilities here in SAA (`getImageWidth.js` and `transformImage.js`)
https://github.com/SU-SWS/saa_alumni/tree/main/src/utilities
Note: you'll only see the downsized image in the Netlify preview, but not in the development build.

src/components/media/simpleImage.js
```diff
@@ -0,0 +1,27 @@
...
+        {caption && (
+          <figcaption>
+            <RichTextRenderer
+              wysiwyg={caption}
+              className={`su-caption su-text-${captionAlign}`}
```
yvonnetangsu, line 18:
```suggestion
              className={`su-caption su-mt-08em su-text-${captionAlign}`}
```
Adding some top margin to the caption (editing free hand so let's see how that looks - might need adjustment) 😃

sherakama (approved):
This is looking good. You are good to merge when you get @yvonnetangsu's blessing.

yvonnetangsu:
@rebeccahongsf For next week when you come back, if you could just add that optional alt text support, then let's call that done for now. The rest (image processing and adding spacing top/bottom) can wait till next sprint, since you have a lot on your plate and we likely want to ask for some design input from Kerri (eg, caption max width etc) 😄

yvonnetangsu (approved):
GTG - thanks @rebeccahongsf! The build error went away after a merge main - it needed some graphQL work that Shea added last night 😄

PR #39, opened 2021-06-14
ADAPT-2706 ADAPT-2113: Masthead Styles

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Desktop masthead styles
- Mobile masthead styles (exclude the actual mobile main nav and the toggle behavior of the menu button)
- Button that opens up the search overlay without the functionality (just styling)
- Fixup Basic card width (fit 100% to parent container, this matters for cards with really short content)

# Review By (Date)
- When does this need to be reviewed by?

# Criticality
- How critical is this PR on a 1-10 scale? Also see [Severity Assessment](https://stanfordits.atlassian.net/browse/D8CORE-1705).
- E.g., it affects one site, or every site and product?

# Review Tasks

## Setup tasks and/or behavior to test

1. https://deploy-preview-39--stanford-alumni.netlify.app/test/yvonne/test-basic-page-interior-hero/
2. https://deploy-preview-39--stanford-alumni.netlify.app/test/yvonne/test-basic-page-interior-no-hero/
3. Look at both of the above previews, one with a hero banner layered beneath, the other without.
4. Make sure to check all breakpoints carefully, please expand your browser width slowly.
5. Desktop breakpoints of particular interest - 992 (LG), 1200 (XL), 1500 (2XL), 1700 (3XL = 2XL + 2 times the screen margin of 100px on each side).

## Front End Validation
- [ ] Is the markup using the appropriate semantic tags and passes HTML validation?
- [ ] Cross-browser testing has been performed?
- [ ] Automated accessibility scans performed?
- [ ] Manual accessibility tests performed?
- [ ] Design is approved by @ user?

### Code
- [ ] Are the naming conventions following our standards?
- [ ] Does the code have sufficient inline comments?
- [ ] Is there anything in this code that would be hidden or hard to discover through the UI?
- [ ] Are there any [code smells](https://blog.codinghorror.com/code-smells/)?
- [ ] Are tests provided? eg (unit, behat, or codeception)

### Code security
- [ ] Are all [forms properly sanitized](https://www.drupal.org/docs/8/security/drupal-8-sanitizing-output)?
- [ ] Any obvious [security flaws or new areas for attack](https://www.drupal.org/docs/8/security)?

## General
- [ ] Is there anything included in this PR that is not related to the problem it is trying to solve?
- [ ] Is the approach to the problem appropriate?

# Affected Projects or Products
- Does this PR impact any particular projects, products, or modules?

# Associated Issues and/or People
- JIRA ticket(s)
[...]

PR #50, opened 2021-06-25
ADAPT-000: GraphQL Build errors (#39)

sherakama:
# READY FOR REVIEW

Fixes build errors caused by no content of type.

RE: https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/#fixing-field-types

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -62,7 +62,7 @@ module.exports = {
         accessToken: process.env.GATSBY_STORYBLOK_ACCESS_TOKEN,
         homeSlug: "home",
         resolveRelations: storyblokRelations,
-        version: process.env.NODE_ENV === "production" ? "published" : "draft", // show only published on the front end site
+        version: "published",
         // version: 'draft'  // would show any including drafts
       },
     },
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -142,3 +142,22 @@ exports.onCreateWebpackConfig = ({
     });
   }
 };
+
+/**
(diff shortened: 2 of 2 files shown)


sherakama:
@yvonnetangsu 

The build errors are caused when no content of something with a graphql query exists and gatsby cannot infer the field types. This will patch up the alerts and the redirects so that if there is no content (published or at all) it will still complete.

gatsby-config.js
```diff
@@ -62,7 +62,7 @@ module.exports = {
         accessToken: process.env.GATSBY_STORYBLOK_ACCESS_TOKEN,
         homeSlug: "home",
         resolveRelations: storyblokRelations,
-        version: process.env.NODE_ENV === "production" ? "published" : "draft", // show only published on the front end site
+        version: "published",
```
sherakama, line 65: When using a public api key and this logic, the build fails even with the graphql type fields added. I figured until we get into workflows and drafts we can just rely on the API key access to account for what we see.

yvonnetangsu: Thank you for investigating this, @sherakama !

yvonnetangsu (approved):
Works well for me. Thank you!

yvonnetangsu:
@sherakama Are the global alerts live on Online Giving? I'm thinking we probably need this fix over there too.

sherakama:
> @sherakama Are the global alerts live on Online Giving? I'm thinking we probably need this fix over there too.

Yep, going to copy this over there too.

PR #53, opened 2021-06-29
ADAPT-2734 Add Rich Media Component

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Add Rich Media Component

# Review By (Date)
- July 6

# Review Tasks
1. Check out branch
2. Run local dev and review `test/rebecca/rich-media-test` or [Netlify Preview](https://deploy-preview-53--stanford-alumni.netlify.app/test/rebecca/rich-media-test/)
## Front End Validation

### Code

### Code security

## General

- [ADAPT-2734](https://stanfordits.atlassian.net/browse/ADAPT-2734)

### DIFF
--- a/src/components/media/embedVideo.js
+++ b/src/components/media/embedVideo.js
@@ -1,0 +1,87 @@
+import React from "react";
+import SbEditable from "storyblok-react";
+import ReactPlayer from "react-player";
+import RichTextField from "../../utilities/richTextRenderer";
+// import CenteredContainer from "../partials/";
+// import FlexCell from "../partials/flexCell";
+
+// Wrapper that sets the size of the video depending on Storyblok option selected
+const VideoWrapper = (
+  props,
+  { blok: { spacingTop, spacingBottom, videoWidth }, children }
+) => {
+  const videoWrapperClasses = `video-embed
+        ${spacingTop !== "none" ? `su-pt-${spacingTop}` : ""}
+        ${spacingBottom !== "none" ? `su-pb-${spacingBottom}` : ""}`;
+
+  // if (videoWidth === "site") {
+  //   return (
+  //     <CenteredContainer classes={videoWrapperClasses}>
+  //       {props.children}
+  //     </CenteredContainer>
+  //   );
+  // }
+
+  // if (videoWidth === "story") {
+  //   return (
+  //     <CenteredContainer flex={true} classes={videoWrapperClasses}>
+  //       <FlexCell lg={8} classes={"su-mx-auto"}>
+  //         {props.children}
+  //       </FlexCell>
+  //     </CenteredContainer>
+  //   );
+  // }
+  // if (videoWidth === "inset") {
+  //   return (
+  //     <CenteredContainer flex={true} classes={videoWrapperClasses}>
+  //       <FlexCell sm={10} md={8} lg={7} xl={6} classes={"su-mx-auto"}>
+  //         {props.children}
+  //       </FlexCell>
+  //     </CenteredContainer>
+  //   );
+  // }
+
+  // This is for fitting to any parent container width so we don't want centered container
+  return <div className={videoWrapperClasses}>{children}</div>;
+};
+
+const EmbedVideo = (
+  props,
+  { blok: { aspectRatio, videoUrl, startMinute, startSecond, caption }, blok }
+) => {
+  const startMin = startMinute ? parseInt(startMinute) : 0;
+  const startSec = startSecond ? parseInt(startSecond) : 0;
+
+  const convertToSecond = (min, sec) => min * 60 + sec;
+
+  return (
+    <SbEditable content={blok}>
+      <VideoWrapper {...props}>
+        <figure className="su-media">
+          <div
+            className={`su-media__wrapper su-embed-container--${aspectRatio}`}
+          >
+            <ReactPlayer
+              url={videoUrl}
+              controls={true}
+              config={{
+                youtube: {
+                  playerVars: { start: convertToSecond(startMin, startSec) },
+                },
+              }}
+            />
+          </div>
+          {caption && (
+            <figcaption
+              className={`su-media__caption ood-story-media__caption su-text-align-${captionAlign}`}
+            >
+              <RichTextField data={caption} />
+            </figcaption>
+          )}
+        </figure>
+      </VideoWrapper>
+    </SbEditable>
+  );
+};
+
+export default EmbedVideo;
--- a/src/components/media/embedVideo.js
+++ b/src/components/media/embedVideo.js
@@ -6,6 +6,7 @@ import RichTextRenderer from "../../utilities/richTextRenderer";
 import {
   smallPaddingBottom,
   smallPaddingTop,
+  mediaAspectRatio,
 } from "../../utilities/dataSource";
 
 // Wrapper that sets the size of the video depending on Storyblok option selected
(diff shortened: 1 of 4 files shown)


src/components/media/embedVideo.js
```diff
@@ -0,0 +1,81 @@
...
+    startMinute,
+    startSecond,
+    caption,
+    aspectRatio,
+    captionAlign,
+  },
+  blok,
+}) => {
+  const startMin = startMinute ? parseInt(startMinute) : 0;
+  const startSec = startSecond ? parseInt(startSecond) : 0;
```
sherakama, line 47: Missing radix param. `parseInt(startMinute, 10)`

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const convertToSecond = (min, sec) => min * 60 + sec;
+
+  return (
+    <SbEditable content={blok}>
+      <VideoWrapper {...blok}>
```
sherakama, line 53: Doesn't blok have a number of properties on it like `_uid` and such that shouldn't be HTML props?

sherakama (changes requested):
Looks good. Thanks for porting this over. Just a couple of comments and please clean up the linting noise.

src/components/media/embedVideo.js
rebeccahongsf: Ack! Great catch @sherakama 😅  I restructured the destructuring of props to the Video Wrapper components. Please do not hesitate to let me know if there's a better way to go about this. Thanks!

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,87 @@
...
+  const startSec = startSecond ? parseInt(startSecond, 10) : 0;
+
+  const convertToSecond = (min, sec) => min * 60 + sec;
+
+  return (
+    <SbEditable content={blok}>
+      <VideoWrapper blok={{ spacingTop, spacingBottom, videoWidth }}>
+        <figure className="su-media">
+          <ReactPlayer
+            className={`su-media__wrapper ${aspectRatio}`}
```
yvonnetangsu, line 62: I totally owe you a review of this one. Looking great - i like how you were using the TW classes. Let's keep that but also adding the aspect ratios to the `dataSource.js`, e.g., in Storyblok, we could add the key-value pair like this instead of having the TW classnames in the value:
key: 16 x 9 (Default)
value: 16x9

And then in dataSource.js
we can add an object with 
`{"16x9": "su-aspect-w-16 su-aspect-h-9"}` 
etc

We're keeping all the classes names inside our code and not on Storyblok, so the JIT mode knows what classes to generate always 😄

sherakama (approved):
GTG by me.

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,90 @@
...
+
+  if (videoWidth === "inset") {
+    return (
+      <Container width="full" className={videoWrapperClasses}>
+        <FlexCell sm={10} md={8} lg={7} xl={6} className="su-mx-auto">
+          {children}
+        </FlexCell>
+      </Container>
+    );
+  }
```
yvonnetangsu, line 33: I think we can simplify this by using the `WIdthBox` component like you did for Simple Image here 😄  
https://github.com/SU-SWS/saa_alumni/blob/main/src/components/media/simpleImage.js

Let's give them these options, plus 6 of 12 columns:
Screen Shot 2021-06-30 at 5 05 48 PM

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,90 @@
...
+} from "../../utilities/dataSource";
+
+// Wrapper that sets the size of the video depending on Storyblok option selected
+const VideoWrapper = ({
+  blok: { spacingTop, spacingBottom, videoWidth },
+  children,
+}) => {
+  const videoWrapperClasses = `video-embed
+        ${spacingTop !== "none" ? smallPaddingTop[spacingTop] : ""}
+        ${spacingBottom !== "none" ? smallPaddingBottom[spacingBottom] : ""}`;
```
yvonnetangsu, line 19: This works, but since we'll be reusing a lot of code on this site. Let's make components more consistent. We were figuring things out on Giving so things were slightly messier there, but since we've figured out what we liked now, let's stick to it 😺 

Could we structure it so that we use separate variables, and our good friend `dcnb` to chain them?
  const paddingTop = smallPaddingTop[spacingTop] ?? ""
  const paddingBottom = smallPaddingBottom[spacingBottom] ?? ""

Here's an example where we use this approach:
https://github.com/SU-SWS/saa_alumni/blob/main/src/components/layout/section.js

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const convertToSecond = (min, sec) => min * 60 + sec;
+
+  return (
+    <SbEditable content={blok}>
+      <VideoWrapper blok={{ spacingTop, spacingBottom, videoWidth }}>
+        <figure className="su-media">
+          <ReactPlayer
+            className={`su-media__wrapper ${
+              mediaAspectRatio !== "none" ? mediaAspectRatio[aspectRatio] : ""
```
yvonnetangsu, line 64: Similarly please use suggested approach above when pulling from dataSource 🙏🏼 😄 
Also, let's default to "16x9" for this one rather than no aspect ratio.

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,90 @@
...
+              youtube: {
+                playerVars: { start: convertToSecond(startMin, startSec) },
+              },
+            }}
+          />
+          {caption && (
+            <figcaption>
+              <RichTextRenderer
+                wysiwyg={caption}
+                className={`su-caption su-mt-06em children:su-leading-snug su-text-${captionAlign}`}
```
yvonnetangsu, line 80: `dcnb` please - and remember JIT mode doesn't like class names concat this way (`su-text-${captionAlign}`) , Check if dataSource.js has a text-align option (you might need to add the right align), then you could use the similar approach as above also 🙏🏼

src/utilities/dataSource.js
```diff
@@ -178,3 +178,9 @@ export const textAlign = {
   center: "su-text-center",
   right: "su-text-right",
 };
+
+export const mediaAspectRatio = {
+  "16x9": "su-aspect-w-16 su-aspect-h-9",
+  "4x3": "su-aspect-w-4 su-aspect-h-3",
+  "1x1": "su-aspect-w-1 su-aspect-h-1",
+};
```
yvonnetangsu, line 186: Perfect 💯  😄

yvonnetangsu (changes requested):
Just a few comments to use existing component for width, add more width options, and use consistent format with `dcnb` and pulling from `dataSource.js`. Thank you for the great work so far!

src/components/media/embedVideo.js
yvonnetangsu: Let's rename "Edge-to-edge" (only the key part) with "Fit parent container" here. Thanks!

yvonnetangsu (approved):
Thank you for the changes - looks great. GTG 🚀

src/components/media/embedVideo.js
```diff
@@ -0,0 +1,87 @@
...
+    videoWidth,
+  },
+  blok,
+}) => {
+  const startMin = startMinute ? parseInt(startMinute, 10) : 0;
+  const startSec = startSecond ? parseInt(startSecond, 10) : 0;
+
+  const convertToSecond = (min, sec) => min * 60 + sec;
+  const aspectRatioStyle = mediaAspectRatio[aspectRatio ?? "16x9"];
+  const captionAlignment = textAlign[captionAlign ?? "left"];
```
yvonnetangsu, line 52: I like your formatting here 😄  ❤️

PR #61, opened 2021-07-02
ADAPT-2839: Remove headlessUi transition and set children to static

Lyafel:
# NOT READY FOR REVIEW

# Summary
- TL;DR - what's this PR for?

# Review By (Date)

# Criticality

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources

### DIFF
--- a/src/components/navigation/mainMenuGroup.js
+++ b/src/components/navigation/mainMenuGroup.js
@@ -1,5 +1,5 @@
-import React, { Fragment } from "react";
-import { Popover, Transition } from "@headlessui/react";
+import React  from "react";
+import { Popover } from "@headlessui/react";
 import { ChevronDownIcon } from "@heroicons/react/solid";
 import CreateBloks from "../../utilities/createBloks";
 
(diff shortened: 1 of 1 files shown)


Lyafel:
@yvonnetangsu Not sure if removing the transition is an acceptable solution but I managed to make it work with removing transition and setting the Panel to static. Is that fine?

yvonnetangsu:
> @yvonnetangsu Not sure if removing the transition is an acceptable solution but I managed to make it work with removing transition and setting the Panel to static. Is that fine?

Oh great! Yeah, I think not having the transition is fine. However, I did find that with this change, when you use the keyboard to navigate and as you tab from link to link inside an expanded dropdown menu, it doesn't automatically close when you tab away from the last link within (currently the dropdown close as you tab away from the last link item in a dropdown). If you could find a way to fix that, that'd be great.

sherakama:
Screen Shot 2021-07-05 at 11 34 02 PM

I think removing the transition is fine but the keyboard navigation and the collapsing when moving to different menu items still needs to happen.

Lyafel:
Right, missed this. I played with it a bit more and couldn't find a way to fix both things at the same time. probably makes sense to go with the second option and remove the library in this case @yvonnetangsu ?

yvonnetangsu:
> Right, missed this. I played with it a bit more and couldn't find a way to fix both things at the same time. probably makes sense to go with the second option and remove the library in this case @yvonnetangsu ?

Thanks for testing, @Lyafel !

@sherakama What to you think? I remember you mentioned that the clicking twice to close issue isn't that critical and we'll just keep headlessUI (and perhaps the TW team will come up with a solution). If not, we can take it out in August since that counts as a bug fix and isn't a new feature.

sherakama:
I just tried to reproduce the issue again and can only get the double click to close to happen on Safari now. Let's lean on Tailwind's headless UI team to address it at this point. We have some time before go-live to wait and see.

yvonnetangsu:
> I just tried to reproduce the issue again and can only get the double click to close to happen on Safari now. Let's lean on Tailwind's headless UI team to address it at this point. We have some time before go-live to wait and see.

Sounds good to me. Let's leave it as is for now then.

PR #62, opened 2021-07-02
ADAPT-2932 Add spacing options and full viewport style

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Add spacing options and full viewport style

# Review By (Date)
- July 7

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

## Front End Validation

### Code

### Code security

## General

- [ADAPT-2932](https://stanfordits.atlassian.net/browse/ADAPT-2932)

### DIFF
--- a/src/components/media/simpleImage.js
+++ b/src/components/media/simpleImage.js
@@ -58,7 +58,7 @@ const SimpleImage = ({
             <img
               src={filename}
               alt={alt ?? ""}
-              className={dcnb(imageStyle, imageFocusPosition)}
+              className={dcnb("su-w-full", imageStyle, imageFocusPosition)}
             />
           </div>
           {caption && (
--- a/src/components/media/simpleImage.js
+++ b/src/components/media/simpleImage.js
@@ -32,9 +32,13 @@ const SimpleImage = ({
 
   let wrapperHeight = "";
   let imageStyle = "";
-  if (imageWidth === "edge-to-edge" && isBanner) {
+  if (imageWidth === "edge-to-edge") {
     wrapperHeight = "su-relative su-h-screen su-w-full su-overflow-hidden";
     imageStyle = "su-h-full su-w-full su-object-cover";
+
+    if (isBanner) {
+      wrapperHeight = "su-h-[30vw] su-relative su-overflow-hidden";
+    }
   }
 
   return (
--- a/src/components/media/simpleImage.js
+++ b/src/components/media/simpleImage.js
@@ -1,27 +1,71 @@
 import React from "react";
 import SbEditable from "storyblok-react";
+import { dcnb } from "cnbuilder";
 import WidthBox from "../layout/widthBox";
 import RichTextRenderer from "../../utilities/richTextRenderer";
+import {
+  smallPaddingBottom,
+  smallPaddingTop,
+  textAlign,
+  objectPosition,
+  containerAlign,
+} from "../../utilities/dataSource";
 
 const SimpleImage = ({
-  blok: { image: { filename, alt } = {}, imageWidth, caption, captionAlign },
+  blok: {
+    image: { filename, alt } = {},
+    imageWidth,
+    imageFocus,
+    caption,
+    captionAlign,
+    spacingTop,
+    spacingBottom,
+  },
   blok,
-}) => (
-  <SbEditable content={blok}>
-    <WidthBox width={imageWidth}>
-      <figure>
-        <img src={filename} alt={alt ?? ""} />
-        {caption && (
-          <figcaption>
-            <RichTextRenderer
-              wysiwyg={caption}
-              className={`su-caption su-mt-06em children:su-leading-snug su-text-${captionAlign}`}
+}) => {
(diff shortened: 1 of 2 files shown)


src/components/media/simpleImage.js
```diff
@@ -1,27 +1,72 @@
...
+  containerAlign,
+} from "../../utilities/dataSource";
 
 const SimpleImage = ({
-  blok: { image: { filename, alt } = {}, imageWidth, caption, captionAlign },
+  blok: {
+    image: { filename, alt } = {},
+    imageWidth,
+    imageFocus,
+    isBanner,
```
rebeccahongsf, line 19: @yvonnetangsu what are our thoughts on having a boolean value to have the full width/height banner? 🤔  That way we don't need to create a new edge-to-edge class and adjust the widthBox.

yvonnetangsu: @rebeccahongsf Nice idea - I like it! 👏🏼  😍

sherakama (approved):
This looks fine to me with @yvonnetangsu's approval.

src/components/media/simpleImage.js
```diff
@@ -1,27 +1,72 @@
...
+  const spacingTopStyle = smallPaddingTop[spacingTop];
+  const spacingBottomStyle = smallPaddingBottom[spacingBottom];
+  const captionAlignment = textAlign[captionAlign ?? "left"];
+  const containerAlignment = containerAlign[captionAlign ?? "left"];
+  const imageFocusPosition = objectPosition[imageFocus ?? "center"];
+
+  let wrapperHeight = "";
+  let imageStyle = "";
+  if (imageWidth === "edge-to-edge" && isBanner) {
+    wrapperHeight = "su-relative su-h-screen su-w-full su-overflow-hidden";
```
yvonnetangsu, line 36: I like the boolean isBanner option, but on Giving, they usually use it as a narrower "band" of image (kind of like a poster without words). Could we copy over the heights of the full width image on Giving for different breakpoints? I think it was 500px or 600px tall for 2XL 😄

yvonnetangsu: Here's an example on this page (2nd image)
https://giving.stanford.edu/stories/academic-and-creative-journey/

yvonnetangsu: Also, could we add a boolean to add a `su-cc` centered container around the caption content, so when the image is full width, one can choose whether the caption start at the edge of of the parent container (eg, when it's placed inside the "main content" and the image is fit to 100% width of parent) or it start at the edge of the centered container (eg, when it's used as an edge to edge image on a page). Thanks 🙏🏼 😄

src/components/media/simpleImage.js
```diff
@@ -1,27 +1,72 @@
...
+      <WidthBox
+        width={imageWidth}
+        className={dcnb(spacingTopStyle, spacingBottomStyle)}
+      >
+        <figure>
+          <div className={wrapperHeight}>
+            <img
+              src={filename}
+              alt={alt ?? ""}
+              className={dcnb(imageStyle, imageFocusPosition)}
```
yvonnetangsu, line 51: Let's add a `su-w-full` class to the image so its width will fill what it's supposed to occupy. I merged this into my branch and found that if the image is smaller than the width you choose, it stays at the smaller size which could be a little bit confusing 😃

yvonnetangsu: Screenshot showing what I was talking about:

Screen Shot 2021-07-06 at 1 26 31 AM

src/components/media/simpleImage.js
rebeccahongsf: I added a `isCaptionCenter` boolean and accompanying help text "If Fit Parent Container is selected, caption can be adjusted to full width or center container aligned." which might need additional revision/massaging. 😄

yvonnetangsu:
Hehe @rebeccahongsf 99% there! Thank you for the changes 😄 

The only issue left is with the Fit Parent Container (non-banner) option, eg, when on mobile breakpoints, the width of the image should fit the width of the device with the height being auto to retain the original image aspect ratio. Currently on mobile:
Screen Shot 2021-07-06 at 4 26 10 PM

src/components/media/simpleImage.js
yvonnetangsu: I think the `su-h-screen` class is causing the issue. I think you can just remove it - we shouldn't need it for any options 😄

rebeccahongsf:
> Hehe @rebeccahongsf 99% there! Thank you for the changes 😄
> 
> The only issue left is with the Fit Parent Container (non-banner) option, eg, when on mobile breakpoints, the width of the image should fit the width of the device with the height being auto to retain the original image aspect ratio. Currently on mobile:
> _screenshot_

Removing the `su-h-screen` seemed to resolve the mobile height issue! Please let me know if anything else appears out of place. Thanks @yvonnetangsu! 😄

yvonnetangsu (approved):
GTG - thank you for the revision! 🎉 😄

PR #67, opened 2021-07-06
ADAPT-2705 Create CtaCard

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Create CtaCard

# Review By (Date)
- July 13

# Criticality

# Review Tasks

1. Check out this branch
2. Review locally at `test/rebecca/cta-card ` or see [Netlify preview](https://deploy-preview-67--stanford-alumni.netlify.app/test/rebecca/cta-card/)

## Front End Validation

### Code

### Code security

## General

- [ADAPT-2705](https://stanfordits.atlassian.net/browse/ADAPT-2705)


src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,26 @@
...
+import CreateBloks from "../../utilities/createBloks";
+
+const CtaCard = ({ blok: { headline, cta }, blok }) => (
+  <SbEditable content={blok}>
+    <div
+      className={dcnb(
+        "cta-card su-w-full su-bg-digital-red su-text-white hocus:su-bg-cardinal-red-xdark su-basefont-23 su-break-words su-pb-50 su-px-50 su-h-[50rem] su-flex su-flex-col su-justify-end"
+      )}
+    >
+      <Heading level="4" font="sans" weight="bold" className={dcnb("su-mb-0")}>
```
yvonnetangsu, line 14: Ahh that reminds me, please add a "Heading level" option in storyblok as well. We have that for all the other cards so might as well add it here :D (please see the basic card for example - there's a datasource for heading level that you could pull from). Thanks!!

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,26 @@
...
+        {headline}
+      </Heading>
+      {cta && (
+        <div className="su-rs-mt-2">
+          <CreateBloks blokSection={cta} />
+        </div>
+      )}
```
yvonnetangsu, line 21: Instead of having this be a nested CTA link. Let's just add a Storyblok Link component (so they can pick internal/external links), and use the `.stretched-link` class to make the whole card a link 😄  (see event card for example)

yvonnetangsu: In the event card, we wrap the SbLink around the headline, but for this CTA card, it probably makes more sense to wrap the link around the CTA text instead, and render the clickable area to cover the whole card 😸

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // Heroicon option
+    const linkIcon = heroicon[icon] ?? heroicon["arrow-right"];
+
+    return (
+      <SbEditable content={blok}>
+        <div
+          className={dcnb(
+            "cta-card su-w-full su-relative su-bg-digital-red su-text-white hocus:su-bg-cardinal-red-xdark su-basefont-23 su-break-words su-p-50 su-min-h-[50rem] su-h-full su-flex su-flex-col su-justify-end"
```
rebeccahongsf, line 48: @yvonnetangsu @sherakama [the ticket ](https://stanfordits.atlassian.net/browse/ADAPT-2705)specified "no max height" so I've set the min height to 500px and height to 100%. With the height set to 100%, if there are multiple cards in a grid row, they will all take on the height of the tallest card. 

Was that the preferred display or did we want them to be staggered?

yvonnetangsu: I'm going to pull this down and test it out in the grid component. We have an option in the grid to make them the same height, so I'm thinking ideally they can be staggered unless that option is checked 😄  Go take your long weekend. I'll review later.

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,89 @@
...
+          )}
+        >
+          <Heading
+            level={headingLevel ?? "3"}
+            font="sans"
+            weight="bold"
+            className={dcnb(
+              "su-mb-0 su-text-white hocus:su-text-white hocus:su-no-underline"
+            )}
+          >
```
yvonnetangsu, line 58: We need to change the level string input from storyblok to a number due to the Decanter-React refactor, so, `level={parseInt(headingLevel, 10) ?? 3}`

Also, even though we allow for different heading levels, we should set a fixed heading font size, e.g., see event.js (under components/content-type/event/). The heading level choices is for accessibility not for font size change 😄

yvonnetangsu:
To test this out properly, I'd suggest adding some CTA cards to the /test/yvonne/test-perks (?) page, since that's the primary page that will use this type of card 😄

src/components/cards/ctaCard.js
yvonnetangsu: I think we could take out the min-height on the card. For the padding, let's use responsive-spacing, I believe for cards, we've using su-rs-p-2 (responsive-spacing step 2). Please check with Kerri if she wants a larger top padding compared to the other 3 sides 😄

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,89 @@
...
+            {headline}
+          </Heading>
+          <SbLink
+            ref={ref}
+            link={link}
+            attributes={rel ? { rel } : {}}
+            classes="su-block su-stretched-link su-group su-transition-colors su-no-underline su-underline-offset su-text-white hocus:su-underline hocus:su-text-digital-red-xlight children:hocus:su-text-digital-red su-mt-[4rem]"
```
yvonnetangsu, line 65: Let's also use responsive spacing for the top margin here as well - just eyeballing, it looks like a su-rs-mt-3 or 2 👀

src/components/cards/ctaCard.js
rebeccahongsf: Thanks Yvonne! 😄  I'll check in with Kerri about the top padding this morning and update the PR accordingly!

src/components/cards/ctaCard.js
yvonnetangsu: Please also add the fixed heading font size 😄  - I think a `su-text-m2` looks about right in this case. The headings should all end up having the same font size no matter what level you select.

yvonnetangsu: There's also a `size` prop in the Heading component 😄

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,91 @@
...
+      iconAnimate = dcnb(
+        iconAnimate,
+        "group-hocus:su-translate-x-01em group-hocus:su--translate-y-01em"
+      );
+    } else {
+      iconAnimate = dcnb(iconAnimate, "group-hocus:su-translate-x-02em");
+    }
+
+    // Heroicon option
+    const linkIcon = heroicon[icon] ?? heroicon["arrow-right"];
```
yvonnetangsu, line 42: Instead of letting them pick the icon, please have it display the appropriate icon depending on the `link.linktype` prop :) In the CTA link, we let the content author pick because we offer them choices of additional icons, e.g., video, download etc, but for card headings (and in this case the link text), since we're only giving them 2 choices, let's add it for them automatically. You could see storyCardView.js as an example.

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+      <SbEditable content={blok}>
+        <div
+          className={dcnb(
+            "cta-card su-w-full su-relative su-bg-digital-red su-text-white hocus:su-bg-cardinal-red-xdark su-basefont-23 su-break-words su-p-50 su-flex su-flex-col su-justify-end"
```
yvonnetangsu, line 48:
```suggestion
            "cta-card su-w-full su-relative su-bg-digital-red su-text-white hocus:su-bg-cardinal-red-xdark su-basefont-23 su-break-words su-rs-py-3 su-rs-px-2 su-flex su-flex-col su-justify-end"
```
Let's use responsive-spacing for card padding 😄

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,86 @@
...
+import SbLink from "../../utilities/sbLink";
+
+const CtaCard = React.forwardRef(
+  (
+    {
+      blok: { headline, headingLevel, icon, linkText, link, rel, srText },
+      blok,
+    },
+    ref
+  ) => {
```
yvonnetangsu, line 16: I don't see a rel field in Storyblok for CTA cards so let's remove it 😄

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,86 @@
...
+              "su-mb-0 su-text-white hocus:su-text-white hocus:su-no-underline"
+            )}
+          >
+            {headline}
+          </Heading>
+          {link && (
+            <SbLink
+              ref={ref}
+              link={link}
+              attributes={rel ? { rel } : {}}
```
yvonnetangsu, line 60: I think this can be removed also

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,86 @@
...
+    // Heroicon option
+    let linkIcon = heroicon["arrow-right"];
+
+    if (link.linktype === "url") {
+      iconClasses =
+        "su-h-08em su-w-08em su-ml-4 su--mt-2 su-transform-gpu su-rotate-45 group-hocus:su-rotate-45";
+      iconAnimate =
+        "su-transition-transform group-hocus:su-transform-gpu group-hocus:su-translate-x-01em group-hocus:su--translate-y-01em";
+      linkIcon = heroicon["external"];
+    }
```
yvonnetangsu, line 36: Thanks! 🙏🏼

src/components/cards/ctaCard.js
yvonnetangsu: We probably don't need the forwardRef and the ref, but it doesn't hurt so no need to remove 😄 

We added the ref in CtaLink because we needed it for creating a focus trap in the search modal, but who knows, there is always a chance a CTA card will be put inside a modal 🤷🏼

yvonnetangsu: Also, the icon prop can now be removed too.

yvonnetangsu (changes requested):
Final batch of feedback left - if you could please remove the `rel` and `icon` prop (and the logic around the `Icon` element since there isn't an `icon` prop anymore, then we're good to go. Thank you - looks great 🙏🏼 😄

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,86 @@
...
+
+    return (
+      <SbEditable content={blok}>
+        <div
+          className={dcnb(
+            "cta-card su-w-full su-relative su-bg-digital-red su-text-white hocus:su-bg-cardinal-red-xdark su-basefont-23 su-break-words su-rs-py-3 su-rs-px-2 su-flex su-flex-col su-justify-end"
```
yvonnetangsu, line 42:
```suggestion
            "cta-card su-w-full su-relative su-transition-colors su-bg-digital-red su-text-white hocus:su-bg-cardinal-red-xdark su-basefont-23 su-break-words su-rs-py-3 su-rs-px-2 su-flex su-flex-col su-justify-end"
```
One more - for more gradual transition of background colors on hover

src/components/cards/ctaCard.js
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import SbEditable from "storyblok-react";
+import Icon from "react-hero-icon";
+import { Heading, SrOnlyText } from "decanter-react";
+import { dcnb } from "cnbuilder";
+import { heroicon, ctaLinkIconColor } from "../../utilities/dataSource";
+import SbLink from "../../utilities/sbLink";
+
+const CtaCard = React.forwardRef(
```
sherakama, line 9: Why the forward ref here?

sherakama (approved):
GTG with Yvonne's approval.

src/components/cards/ctaCard.js
yvonnetangsu: I think Rebecca might have modeled this after our CtaLink component which has the forwardRef (Yanik added the ref for focus trapping within the search modal since there were CTA links in the modal).

sherakama: Let's remove it until we need it

yvonnetangsu (approved):
Thank you for all the revisions - nice job, @rebeccahongsf ! I just ran a final eslint - GTG 🥳

PR #81, opened 2021-07-16
Update htmlHead.js

sherakama:
# READY FOR REVIEW

Fixes build error/warning

```
error Warning: Received `true` for a non-boolean attribute `crossOrigin`.

If you want to write it to the DOM, pass a string instead: crossOrigin="true" or crossOrigin={value.toString()}.
    in link
    in Fragment
    in HtmlHead
    in head
    in html
    in HTML
```


yvonnetangsu (approved):
Thank you for fixing this!

yvonnetangsu:
I guess eslint doesn't like this one either huh

PR #90, opened 2021-07-20
Delete _redirects

sherakama:
# READY FOR REVIEW

# Summary
- Delete redirects file.
- We don't need these here. The `/cdn` URLs are legacy for the first version of the asset masking and are not required for these sites.


yvonnetangsu (approved):
GTG. Thanks, @sherakama!

PR #93, opened 2021-07-22
ADAPT-3028: Sanitized IDs for WYSIWYG, posters, sections; update form elements to use new version of decanter; remove gatsby-plugin-anchor-links

yvonnetangsu:
# READY FOR REVIEW

# Summary
- This PR involved slightly more things than I originally anticipated.
- Enable id field for individual WYSIWYG and poster components, since Storyblok doesn't allow you to add an anchor in the WYSIWYG content. This would allow anchor links to link to a WYSIWYG or a poster on the page.
- I recall Nathan removed gatsby-plugin-anchor-links since that interferes with anchor links working properly on the Giving site, so I removed it
- When updating the plugins, I did a npm package refresh and ended up pulling in a new version of Decanter. Thus, I have to update the path to the input styles in the tailwind.config.js file
- Since Decanter v7 latest version doesn't have button base styles, I added some of the styles back to the buttons on this site - hamburger, main nav desktop parent items, search buttons (the one that opens up the modal, and the submit search button inside the modal)
- Update: Use React-id-generator to sanitize ids added by content authors (on Sections, Posters and WYSIWYG)

# Review By (Date)

# Criticality

# Review Tasks

1. Go to https://deploy-preview-93--stanford-alumni.netlify.app/test/yvonne/test-basic-page-interior-no-hero/
2. Check that this WYSIWYG in the main content region with the heading "Heading for anchor link" has an id "test"
Screen Shot 2021-07-22 at 5 08 30 PM

3. Scroll down and check that the red poster has the id "red-poster"
Screen Shot 2021-07-22 at 5 08 59 PM

4. Check the main nav parent item has white text
5. Check that the hamburger button has white icon (regular non hover/focus state)
6. Check that the search button in the masthead has white text
7. Click on it to open up the modal, check that the red submit search button will hover/focus to a soft black color.

## Front End Validation

### Code

### Code security

## General

@rebeccahongsf 
@Lyafel 
- JIRA ticket(s)
- Other PRs

# Resources


tailwind.config.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
     require(dir + "/base/base.js")(),
 
     // @tailwind components;
-    require(decanter + "/components/form/input-base.js")(),
+    require(decanter + "/components/form/input.js")(),
```
yvonnetangsu, line 60: Updated the path to point to the file that was renamed in recent decanter PR

package.json
```diff
@@ -13,12 +13,11 @@
     "@heroicons/react": "^1.0.1",
     "@tailwindcss/aspect-ratio": "^0.2.1",
     "algoliasearch": "^4.9.3",
-    "decanter-react": "^1.0.7-alpha.0",
+    "decanter-react": "^1.0.7",
     "gatsby": "^3.4.2",
-    "gatsby-plugin-anchor-links": "^1.2.1",
```
yvonnetangsu, line 18: Removed this as this plugin was flagged by Nathan that it actually broke anchor linking in Giving.

yvonnetangsu:
FYI @sherakama After this is merge, @rebeccahongsf  has a ticket to verify the anchor links so I'm not doing additional work on anchor links for now other than removing the problematic plugin `gatsby-plugin-anchor-links` that Nathan flagged.

yvonnetangsu:
@Lyafel Heads up this PR will include some changes to the button, since some of the base styles has been removed in Decanter v7 - I'm adding back some of the styles using TW classes into our component but please double check after this get merged in since there might be some search button component that I missed.

yvonnetangsu:
@sherakama If you could please take a look at this after your meetings today, that'd be awesome. This pulls in a newer version of Decanter so it would be easier if it gets into main. Thank you 🙏🏼

tailwind.config.js
sherakama: I think you are going to have to add the `form-elements` file as well. The checkboxes on the search page are unusable at the moment. 

Screen Shot 2021-07-23 at 10 09 01 AM

src/components/composite/poster.js
```diff
@@ -83,6 +84,7 @@ const Poster = ({
           bgFilename,
           "linear-gradient(240deg, rgba(24, 29, 28) 10%, rgba(98, 0, 89, 0.85) 60%, rgb(177, 4, 14) 100%)"
         )}
+        id={id}
```
sherakama, line 87: This prop is coming from user input. Please protect our hardcoded IDs by sanitizing out the anchors we have hardcoded like `#page-title` and ensure uniqueness by wrapping the value in something like `react-id-generator` https://github.com/SU-SWS/ood_online_giving/blob/main/src/components/embed/giveGabForm.js#L12

src/components/simple/wysiwyg.js
```diff
@@ -20,6 +20,7 @@ const Wysiwyg = ({
       <WidthBox
         width={width ?? "edge-to-edge"}
         className={dcnb(mySpacingTop, mySpacingBottom)}
+        id={id}
```
sherakama, line 23: This is user input. Please sanitize and protect our hardcoded ids.

sherakama (changes requested):
Thanks for getting to this upgrade but a few changes are required.

tailwind.config.js
yvonnetangsu: Thanks for that! Updated

yvonnetangsu: Looks like I'll have to add the su- classnames to them as well

yvonnetangsu: This is done. I'll go work on sanitizing the id

yvonnetangsu:
@sherakama for next week:

A randomly generated 4 digital integer is attached to each page's main content id (eg, #main-content-8462) and page title id,. The skiplink, and hero scroll to link (when there is a hero component on the page) have both been updated to include this random integer generated on each page reload.

Please check a basic page, a dark page, and a story page; and test the skip link and hero banner scroll to link. Story page does not allow for adding a hero component so just check the skip link for story page is enough.

Basic page with hero
https://deploy-preview-93--stanford-alumni.netlify.app/test/yvonne/test-basic-page-interior-hero/

Dark page with hero
https://deploy-preview-93--stanford-alumni.netlify.app/test/yvonne/test-dark-page-with-hero/

Story page
https://deploy-preview-93--stanford-alumni.netlify.app/stories/back-to-the-land/

src/components/composite/poster.js
sherakama: I'd still like to see this value run through something like the `react-id-generator` to ensure uniqueness. What happens when multiple posters are placed on the same page and the content author adds the same ID?

src/components/content-types/story/storyPageView.js
```diff
@@ -91,10 +93,10 @@ const StoryPageView = (props) => {
 
   return (
     <SbEditable content={blok}>
-      <Layout {...props}>
+      <Layout mainContentId={`main-content-${uniqueIdSuffix}`} {...props}>
         <Container
           element="main"
-          id="main-content"
+          id={`main-content-${uniqueIdSuffix}`}
```
sherakama, line 99: I suppose this works here with the random integers but this is not a solution you would want to use for something like heading anchors or in content anchors as you wouldn't be able to share a URL with the anchor and have it work. Randomizing an anchor's ID limits its use to that specific page load and is not sharable via URL. 

Another approach would have been to run `main-content` and `page-title` through the `react-id-generator` package as that will prevent subsequent use of those ids as long as all IDs go through the generator.

src/components/simple/wysiwyg.js
sherakama: This still hasn't been sanitized to prevent duplicate ids.

src/components/page-types/basicPage.js
```diff
@@ -37,22 +39,29 @@ const BasicPage = (props) => {
...
+              pageTitleId={`page-title-${uniqueIdSuffix}`}
+            />
             <Container>
               <Heading
                 level={1}
                 align="center"
                 font="serif"
                 srOnly={isSrOnlyTitle}
-                id="page-title"
+                id={`page-title-${uniqueIdSuffix}`}
```
sherakama, line 64: Don't we want this to be consistently `page-title` so we don't have to pass props around and guess where to anchor to from the other components?

sherakama (changes requested):
I think you went a little too far with the random integer suffixes. I think you can revert back to having just the `main-content` and `page-title` IDs hardcoded in our templates as long as you run all content author inputted IDs through a unique ID enforcer. It would be best if the anchors did not randomize so that they can be predictable and bookmarkable.

src/components/simple/wysiwyg.js
yvonnetangsu: @sherakama I guess I misunderstood your request. What I did was that I protected our 2 hardcoded ids which are placed on main content (for the skiplink) and the other on the page title (for the hero banner link). I added the random numbers at the end so it's unlikely that the users will add the same ids (since they're now randomized) to posters and WYSIWYGs.

I see now that your requests is to use the react-id-generator on the WYSIWYG and poster IDs as well - my only concern is that, there is no easy way for the content author to figure out the ids to link to unless they inspect it in the inspector (seems like they need to be trained) 🤔

src/components/content-types/story/storyPageView.js
yvonnetangsu: Yeah, the randomized one is only for the 2 anchors that we need for skiplinks and hero banner links (they're not used by content authors, just us in the code).

yvonnetangsu:
> I think you went a little too far with the random integer suffixes. I think you can revert back to having just the `main-content` and `page-title` IDs hardcoded in our templates as long as you run all content author inputted IDs through a unique ID enforcer. It would be best if the anchors did not randomize so that they can be predictable and bookmarkable.

Yeah I think I went a bit far, buuuut to my defense, I only randomized the 2 IDs that won't be used by regular content authors - I don't anticipate anyone would actually link to the `<main>` region (the main-content is for skiplink only) or the page title (which is often visible to screen reader only - the link is for the hero banner arrow link). The reason I added the randomized numbers at the end of these 2 ID is that, now a content author won't be able to add an id to a WYSIWYG or poster that ends up being identical to the page-title or main-content ids, since I suspect they won't be adding a 4 digital number at the end 😂 

I understand your concern with the other ids that the content authors can add, eg, for the posters and WYSIWYG and want to protect them to make them unique. I'll revise to run them through the id generator, but I think we need to add a note to Dena to add that to the training so the content authors will be able to inspect the id that's sanitized and link correctly to that.

src/components/page-types/basicPage.js
yvonnetangsu: Yup, this isn't a prop that's passed in from Storyblok. it's hardcoded page-title plus the random number. I passed it down as a prop from the parent Layout component where it was generated, so the Hero banner link gets the same random number, ie, when someone add a hero banner on the page, it will always link to the page-title (the pair will always have the same random integer at the end).

yvonnetangsu:
> I think you went a little too far with the random integer suffixes. I think you can revert back to having just the `main-content` and `page-title` IDs hardcoded in our templates as long as you run all content author inputted IDs through a unique ID enforcer. It would be best if the anchors did not randomize so that they can be predictable and bookmarkable.

Ok 😄  I removed the overkill stuff, and went with your approach, ran all content author inputted IDs through the react-id-generators (on the Section, WYSIWYG and Poster components), and went back to just main-content and page-title for those 2 hardcoded ids. Ready for another look - thanks, @sherakama 🙏🏼

yvonnetangsu:
And @sherakama Here's a page where you could check both the skiplink and the hero scroll to link
https://deploy-preview-93--stanford-alumni.netlify.app/test/yvonne/test-basic-page-interior-hero/

I updated the page with the anchors in the "steps to test" page as well (the page with no hero banner)

src/components/simple/wysiwyg.js
sherakama: > My only concern is that, there is no easy way for the content author to figure out the ids to link to unless they inspect it in the inspector

I see what you mean. Perhaps nextId is not the right library for this. I am hoping for something that would ensure a unique string but not necessarily append an integer to every instance. For example, I would be looking for a function that I could pass 'test' into and get 'test' back from if it had not been used before. For subsequent inputs of 'test' I would get the added integer. 

This also gets a little nastier with the static build process as that is all one operation, and although having 'test' as an ID one time per page is legal, the fact that the build process is one operation, multiple pages with the name ID will cause the ID to increment. We would have to also nullify the increments on each page context.

I'll create a new ticket for this work.

sherakama: https://stanfordits.atlassian.net/browse/ADAPT-3075

sherakama (approved):
GTG

src/components/simple/wysiwyg.js
yvonnetangsu: Yeah totally agree with you there -  I'm a little worried about the id incrementing also as it is now not very predictable. Also, if I add a section in between 2 older sections, I'm suspecting that the id #s are going to change which would cause any existing anchor links to not work 😬  Thanks for creating that ticket. Let's investigate in August.

PR #107, opened 2021-08-05
ADAPT-3092: Wysiwyg center text option

Lyafel:
# Summary
- Added an option to the wysiwyg component that will add left or centered text class 

# Review By (Date)

# Criticality
- 1

# Review Tasks

1. Check out this branch
2. Navigate to help/stanfordpass 
3. Verify that text is centered now

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources

### DIFF
--- a/src/components/simple/wysiwyg.js
+++ b/src/components/simple/wysiwyg.js
@@ -5,12 +5,13 @@ import nextId from "react-id-generator";
 import {
   smallPaddingBottom,
   smallPaddingTop,
+  textAlign,
 } from "../../utilities/dataSource";
 import RichTextRenderer from "../../utilities/richTextRenderer";
 import WidthBox from "../layout/widthBox";
 
 const Wysiwyg = ({
-  blok: { content, width, spacingTop, spacingBottom, isLightText, textPosition, id },
+  blok: { content, width, spacingTop, spacingBottom, isLightText, align, id },
   blok,
 }) => {
   const wysiwygId = nextId(`${id}-`);
@@ -27,7 +28,10 @@ const Wysiwyg = ({
         <RichTextRenderer
           isDark={isLightText}
           wysiwyg={content}
-          className={`su-heading-link-icon ${textPosition}`}
+          className={dcnb(
+            "su-heading-link-icon",
+            textAlign[align] ?? textAlign.left
+          )}
         />
       </WidthBox>
     </SbEditable>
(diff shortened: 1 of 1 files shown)


src/components/simple/wysiwyg.js
```diff
@@ -10,7 +10,7 @@ import RichTextRenderer from "../../utilities/richTextRenderer";
 import WidthBox from "../layout/widthBox";
 
 const Wysiwyg = ({
-  blok: { content, width, spacingTop, spacingBottom, isLightText, id },
+  blok: { content, width, spacingTop, spacingBottom, isLightText, textPosition, id },
```
yvonnetangsu, line 13: If you could please rename this prop to `textAlign`, that'd be great. Just so it's consistent with other components.

src/components/simple/wysiwyg.js
```diff
@@ -27,7 +27,7 @@ const Wysiwyg = ({
         <RichTextRenderer
           isDark={isLightText}
           wysiwyg={content}
-          className="su-heading-link-icon"
+          className={`su-heading-link-icon ${textPosition}`}
```
yvonnetangsu, line 30: I see that you're adding the Tailwind classnames directly in Storyblok as the values in the key-value pairs. However, the Tailwind JIT mode needs all the TW classes that we'll ever use in our code to know what classes to generate, so having them in Storyblok isn't safe. They might seem like they're working because we have used `su-text-center` elsewhere in the project so the CSS for that classname has already been generated. 

To handle this, we have a `datasource.js` file with variable objects that you can use. One is for text alignment:
https://github.com/SU-SWS/saa_alumni/blob/main/src/utilities/dataSource.js

On the Storyblok side, please change the values in the dropdown key-value pairs from `su-text-left` or `su-text-center` to just `left` and `center`, then you can reference dataSource to retrieve the right classname based on the prop.

yvonnetangsu: Also, please import `dcnb` and use it to chain classnames, thanks 🙏🏼  

Example here:
https://github.com/SU-SWS/saa_alumni/blob/main/src/components/layout/grid.js

yvonnetangsu (changes requested):
Comments left. Please revise - thanks, @Lyafel !

src/components/simple/wysiwyg.js
Lyafel: Looks like the names clash with the textAlign datasource, some other component I found this in used it as just align so that;s what I changed it to, hope it's ok?

yvonnetangsu: Yeah, that's ok, thanks!

yvonnetangsu (approved):
GTG - thanks for the revision

PR #114, opened 2021-08-10
ADAPT-3103: Updated unique ID for embed cards, to avoid error.

nplowman:
# READY FOR REVIEW

# Summary
- Fixes render-breaking error on builds when using Embed Card component.

# Review By (Date)
- By end of sprint.

# Criticality
- 9/10 - Without this fix, white screen of death is seen when using embed card component.

# Review Tasks

1. Visit a page that is using the Embed Card component, such as /test/yvonne/test-home-embed-error/

## Front End Validation

### Code

### Code security

## General

- https://stanfordits.atlassian.net/browse/ADAPT-3103

# Notes
- The issue had something to do with the ID generator library not behaving as expected on static builds.
(I wasn't able to pin it down 100%, but I think what was happening was that it was re-generating the ID during hydration causing a mismatch with what was in the initial server-side render. )
- Moving away from the dynamically-generated ID to a static ID based on the Storyblok UUID resolved the issue.
- I tried duplicating a card to ensure this would not introduce issues with duplicate IDs. It looks like Storyblok creates a new UUID for each instance, so this shouldn't be an issue.

### DIFF
--- a/src/components/cards/embedCard.js
+++ b/src/components/cards/embedCard.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
 /* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable no-restricted-syntax */
 /* eslint-disable react/no-danger */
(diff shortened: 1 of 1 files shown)


sherakama (approved):
LG2M

PR #159, opened 2021-09-08
ADAPT-3255 | trip page layout

mattanglin:
# NOT READY FOR REVIEW

# Summary
- Fixed bug with gatsby creating pages for global components
- Fixed bug with invalid classname property
- Basic trip page layout component.

## Notes
Just left this basic while I exploring existing components in sb and react. Assuming the additional tickets will fill it out.

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -7,12 +7,24 @@ exports.createPages = ({ graphql, actions }) => {
 
   return new Promise((resolve, reject) => {
     const storyblokEntry = path.resolve("src/templates/storyblok-entry.js");
+    // Omit non-page entities from page generation
+    const contentTypesToOmit = [
+      "alert",
+      "localFooter",
+      "masthead",
+      "redirect", // NOTE: Redirects are omitted as they are specifically generated below
+      "searchEntry",
+      "searchKeywordBanner",
+      "searchSuggestions",
+      "verticalNavWrapper",
+    ];
+    const omittedComponentsArray = JSON.stringify(contentTypesToOmit);
 
     resolve(
       graphql(
         `
           {
-            allStoryblokEntry {
+            allStoryblokEntry (filter: { field_component: { nin: ${omittedComponentsArray} } }) {
               edges {
                 node {
                   id
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "@tailwindcss/aspect-ratio": "^0.2.1",
     "@tailwindcss/forms": "^0.3.3",
     "algoliasearch": "^4.9.3",
+    "cnbuilder": "^2.7.1",
     "decanter-react": "^1.0.7",
     "gatsby": "^3.4.2",
     "gatsby-link": "^3.12.0",
--- a/src/components/components.js
+++ b/src/components/components.js
@@ -48,6 +48,7 @@ import Section from "./layout/section";
 import SimpleImage from "./media/simpleImage";
 import Story from "./content-types/story/story";
 import StoryCard from "./cards/storyCard";
+import TripPage from "./page-types/TripPage";
 import VerticalNav from "./navigation/verticalNav";
 import VerticalNavWrapper from "./navigation/verticalNavWrapper";
 import VerticalNavItem from "./navigation/verticalNavItem";
@@ -92,6 +93,7 @@ const ComponentList = {
   perkCard: PerkCard,
   perkCardHorizontal: PerkCardHorizontal,
   poster: Poster,
+  redirect: Redirect,
   searchFacet: SearchFacet,
   searchField: SearchField,
   searchPage: SearchPage,
@@ -102,7 +104,7 @@ const ComponentList = {
   simpleImage: SimpleImage,
   story: Story,
   storyCard: StoryCard,
-  redirect: Redirect,
+  trip: TripPage,
   verticalNav: VerticalNav,
   verticalNavWrapper: VerticalNavWrapper,
   verticalNavItem: VerticalNavItem,
--- a/src/components/page-types/TripPage.jsx
+++ b/src/components/page-types/TripPage.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import SbEditable from "storyblok-react";
+import Layout from "../partials/layout";
+
+const TripPage = (props) => {
+  console.log('TripPage:', props.blok);
+
+  return (
+    <Layout hasHero {...props}>
+      <div>TripPage</div>
+    </Layout>
+  );
+};
+
+export default TripPage;
--- a/src/components/page-types/TripPage.jsx
+++ b/src/components/page-types/TripPage.jsx
@@ -1,14 +1,24 @@
-import React from 'react';
+import React from "react";
 import SbEditable from "storyblok-react";
+import { Container } from "decanter-react";
 import Layout from "../partials/layout";
 
 const TripPage = (props) => {
-  console.log('TripPage:', props.blok);
+  const { blok } = props;
 
   return (
-    <Layout hasHero {...props}>
-      <div>TripPage</div>
-    </Layout>
+    <SbEditable content={blok}>
+      <Layout hasHero {...props}>
+        <Container
+          element="main"
+          id="main-content"
+          className="trip-page su-relative su-flex-grow su-w-full"
+          width="full"
+        >
+          {/* Trip Page bloks here */}
+        </Container>
+      </Layout>
+    </SbEditable>
   );
 };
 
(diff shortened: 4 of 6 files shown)


src/components/search/openSearchModalButton.js
```diff
@@ -1,16 +1,17 @@
...
   <>
     <Button
       type="button"
-      className={{
+      className={dcnb({
         "su-leading-display": false,
         "hocus:su-bg-transparent": false,
         "search-button su-flex-shrink-0 su-w-40 su-h-40 su-ml-auto su-items-center su-justify-content-center lg:su-w-auto lg:su-h-auto lg:su-pl-16 lg:su-pr-13 lg:su-pt-2 su-rounded-full lg:su-text-18 su-text-white su-leading-none su-border-2 su-border-solid su-border-digital-red-xlight su-transition-colors hocus:su-bg-cardinal-red-xxdark xl:hocus:su-bg-cardinal-red-xdark lg:hocus:su-no-underline": true,
-      }}
+      })}
```
yvonnetangsu, line 14: I've found that we don't actually need to write out dcnb here, since the `Button` component is pulled in from Decanter-React, and the Button `className` prop already has dcnb built in so passing the className as an object without the dcnb works ok.

src/components/page-types/TripPage.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import SbEditable from "storyblok-react";
+import { Container } from "decanter-react";
+import Layout from "../partials/layout";
+
+const TripPage = (props) => {
```
yvonnetangsu, line 6: Any reason to use .jsx instead of .js for this flle?

gatsby-node.js
```diff
@@ -7,12 +7,24 @@ exports.createPages = ({ graphql, actions }) => {
...
+      "alert",
+      "localFooter",
+      "masthead",
+      "redirect", // NOTE: Redirects are omitted as they are specifically generated below
+      "searchEntry",
+      "searchKeywordBanner",
+      "searchSuggestions",
+      "verticalNavWrapper",
+    ];
+    const omittedComponentsArray = JSON.stringify(contentTypesToOmit);
```
yvonnetangsu, line 21: Thanks for doing this!

package.json
```diff
@@ -14,6 +14,7 @@
     "@tailwindcss/aspect-ratio": "^0.2.1",
     "@tailwindcss/forms": "^0.3.3",
     "algoliasearch": "^4.9.3",
+    "cnbuilder": "^2.7.1",
```
yvonnetangsu, line 17: We've been able to use dcnb since it's already a dependency of Decanter React, but probably adding it explicitly is a good idea.

src/components/page-types/TripPage.jsx
```diff
@@ -0,0 +1,25 @@
...
+        <Container
+          element="main"
+          id="main-content"
+          className="trip-page su-relative su-flex-grow su-w-full"
+          width="full"
+        >
+          {/* Trip Page bloks here */}
+        </Container>
+      </Layout>
+    </SbEditable>
```
yvonnetangsu, line 21: Looks good to me! 👍🏼

src/components/components.js
```diff
@@ -102,7 +104,7 @@ const ComponentList = {
   simpleImage: SimpleImage,
   story: Story,
   storyCard: StoryCard,
-  redirect: Redirect,
+  trip: TripPage,
```
yvonnetangsu, line 107: We've been make these the same except for the first letter capitalized (ie, using the current convention this would be `trip: Trip`). Any reason to deviate from that? For the Story content type, we make the Story component (story.js) into a switcher between the card view or page view. Are we using a different convention for the Trip content type?

Merely a question 😄  Open to ideas and suggestions for improvements.

yvonnetangsu:
Thanks for the global components change! Left a few questions for you, @mattanglin 😄

gatsby-node.js
yvonnetangsu: We probably want this in production `main` also.

gatsby-node.js
```diff
@@ -7,12 +7,24 @@ exports.createPages = ({ graphql, actions }) => {
...
 
     resolve(
       graphql(
         `
           {
-            allStoryblokEntry {
+            allStoryblokEntry (filter: { field_component: { nin: ${omittedComponentsArray} } }) {
```
sherakama, line 27: This is a nice improvement. Thanks.

package.json
sherakama: It doesn't seem to hurt but I usually leave out calling sub-dependencies in parent packages unless I need a specific version for something. I find it keeps the long-term maintenance down.

yvonnetangsu: Good to know, @sherakama !

src/components/search/openSearchModalButton.js
yvonnetangsu: Just did some testing - looks like this change would cause the dedupe function to not work. On live, it seems to work as intended. This is from this PR's Netlify preview, and it looks like the `hocus:su-bg-transparent` class is still there. 
Screen Shot 2021-09-09 at 9 36 33 AM

package.json
mattanglin: Using modules that aren't explicitly defined in your package.json and relying on them from your submodules can be a dangerous game...

yvonnetangsu: That's a good point too, @mattanglin 😄  I like how we're having a discussion about this. More devs = goood

src/components/page-types/TripPage.jsx
mattanglin: `.jsx` and `.tsx` are standard extensions for files containing React JSX markup. I prefer using the `x` extension variant to distinguish components from utility/service code.

yvonnetangsu: That sounds like a good idea to me. In that case, let's make that a standard for the whole repo and perhaps we can update all the file extensions in one go (could be another ticket).

src/components/search/openSearchModalButton.js
mattanglin: This was throwing an error in the console as `className` is a reserved prop that's expected to be a string.
image


I can now see that Button [accepts className as an object](https://github.com/SU-SWS/decanter-react/blob/main/src/Button/Button.js#L186-L189) and it's getting [passed to dcnb](https://github.com/SU-SWS/decanter-react/blob/main/src/Button/Button.js#L109-L115) in decanter-react.
_IMO,_ the `className` prop shouldn't altered from it's standard.

src/components/components.js
mattanglin: I named the Storyblok component `trip` as that seemed semantically appropriate in the context of Storyblok (and for a content author). I named the React component `TripPage` as that seemed most appropriate in the context of the codebase (I had already ran into some confusion trying to discern `event`/`perk` cards and pages). Trip is an entity that has multiple views (page/card/filtering), but it seemed redundant to create a `tripPage` in Storyblok since every trip generates a page 🤷🏼.

Not sure I completely understand the logic behind having the parent Story switcher component and when, if ever, the [first layout conditional](https://github.com/SU-SWS/saa_alumni/blob/main/src/components/content-types/story/story.js#L6-L8) is hit.

yvonnetangsu: We have a Story Card component, which allow the users to select a story that's added to the site, and display it as a card (similar to the masthead picker). That's where the layout="card" is set:
https://github.com/SU-SWS/saa_alumni/blob/main/src/components/cards/storyCard.js

If the layout prop isn't set, it defaults to use the story page layout.

src/components/search/openSearchModalButton.js
yvonnetangsu: Yeah...about the className. @sherakama and I went back and forth on this when we first started working on Decanter React. We want to make it intuitive for developers to pass down extra classnames as props, in case they want to customized the components they import (which happens often) into their projects. Using "className" probably isn't ideal, but it is a bit more intuitive.
@sherakama tagging you in case you have thoughts on this.

src/components/components.js
yvonnetangsu: Examples of the Story Cards - if you click on them, it brings you to the page layout
https://alumni-preview.stanford.edu/test/yvonne/test-story-cards/

mattanglin: Right. I guess I'm making the argument that this is overcomplicating the rendering logic. The `storyCardView` and `storyPageView` components effectively handle the `story` and `storyCard` Storyblok components. Adding the `story` and `storyCard` react components are completely unnecessary.

yvonnetangsu: Hmm..... let's try it your way then, @mattanglin! I'm up for simplifying things 😃

yvonnetangsu (changes requested):
@mattanglin If you could please reverse the modal button dcnb change (for now at least, until we decide what to do with className prop), then this is GTG for me. Really nice improvement with omitting the page generation for global components. Thanks!

yvonnetangsu (approved):
GTG thank you!

PR #163, opened 2021-09-14
ADAPT-3263 | @mattanglin | trip card

mattanglin:
#  READY FOR REVIEW

# Summary
- Adds trip card component.

# Review Tasks

1. View trip cards at https://6140d02298b8ec009e020da1--stanford-alumni-uat.netlify.app/test/matt/test-trip-card/


src/components/cards/TripCard/TripCard.jsx
```diff
@@ -0,0 +1,84 @@
...
+}) => {
+  // Resolve Card props
+  const tripTitle = cardTitle || title;
+  const tripSubtitle = cardSubtitle || subtitle;
+  const description = cardDescription || shortDescription;
+  const image = cardImage?.filename ? cardImage : heroImage;
+  const tripURL = `/${fullSlug.replace(/^\//, "")}`;
+
+  return (
+    <FlexBox direction="col" element="article" className={styles.root}>
```
mattanglin, line 42: Ideally, this would've been extracted out into a more reusable component for cards, but I stopped here rather than try to refactor all the other cards in this PR.

yvonnetangsu:
Thank you for the work, @mattanglin! I'll review today.

src/components/simple/Date/Date.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { getDate } from "./getDate";
+
+export const DateProps = {
+  startDate: PropTypes.string.isRequired,
+  endDate: PropTypes.string,
+  isMinimal: PropTypes.bool,
+  className: PropTypes.string,
+};
+
+export const Date = ({ startDate, endDate, isMinimal, className }) => {
```
mattanglin, line 14: I saw that we have the `DateBlock` component, but wanted to encapsulate the functionality and make it more developer friendly in terms of composition (and seeing that we were duplicating logic in multiple places).
Once again, there is some amount of duplication due to not wanting to do a complete refactor in this PR.

src/components/cards/TripCard/TripCard.jsx
```diff
@@ -0,0 +1,84 @@
...
+      <div className={styles.imgWrapper}>
+        <CardImage
+          className={styles.image}
+          filename={image.filename}
+          smartFocus={image.focus}
+          loading="lazy"
+          size="vertical_3x2"
```
mattanglin, line 49: I know the design currently has 1x1 images, but also know that there are currently discussions for changing that around. We'll likely need to revisit this at some point in the future.

mattanglin:
> Thank you for the work, @mattanglin! I'll review today.

I coded some things in a slightly different way than what currently exists in the codebase, which I'm happy to discuss my reasoning for and/or revert if it's causing too much friction.

src/types/TripType.js
```diff
@@ -0,0 +1,31 @@
...
+    startDate: PropTypes.string,
+    endDate: PropTypes.string,
+    heroImage: SBImageType,
+    cardTitle: PropTypes.string,
+    cardSubtitle: PropTypes.string,
+    cardDescription: PropTypes.string,
+    tag: CardTagType,
+    // TODO: Add additional SB trip entity field types for comprehensive type checking
+  }),
+};
```
yvonnetangsu, line 31: This is a good idea - PropTypes probably not as good as Typescript but it's the next best thing for now I guess 😂  I'm up for this.

src/components/cards/TripCard/TripCard.jsx
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Link } from "gatsby";
+import { FlexBox, Heading } from "decanter-react";
+import CardImage from "../../media/cardImage";
+import { Date } from "../../simple/Date/Date";
+import HeroIcon from "../../simple/heroIcon";
+import TabLabel from "../../simple/tabLabel";
+import { TripType } from "../../../types/TripType";
+import * as styles from "./TripCard.styles";
```
yvonnetangsu, line 10: I quite like this idea of separating the styles out. Make the jsx cleaner 🤔

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
...
+  "su-group su-relative su-overflow-hidden su-bg-black-true su-break-words su-border-su-border-black";
+export const content = "su-px-26 su-pb-34 su-text-white";
+export const image =
+  "su-w-full su-h-full su-transition-transform su-transform-gpu group-hover:su-scale-[1.03] group-focus-within:su-scale-[1.03]";
+export const date =
+  "su-relative su-mt-[-5.6rem] lg:su-mt-[-6.3rem] su-mb-26 su-z-10";
+export const link =
+  "su-stretched-link su-z-10 su-rs-mt-0 su-mb-08em su-no-underline hocus:su-underline su-underline-offset !su-underline-thick !su-underline-digital-red-xlight";
+export const heading =
+  "su-mt-32 su-type-4 su-font-bold su-relative su-inline su-text-white";
```
yvonnetangsu, line 11: I've found that for card headings I prefer to add the font size `su-type-4` in the link, and add `su-type-0` in the heading. This way it scales the arrow icon up perfectly also.

On a related note, I'm asking Bhavika on Figma if we can scale down the trip card heading to `su-type-3` (so it matches the event card with the "large" heading option, and also makes it a bit smaller since it's quite large now).

yvonnetangsu: Got confirmation from Bhavika that we can step the title font size fo `su-type-3`. So if you add `su-type-3` to `link`, and `su-type-0` to `heading` here, it would make it look the same as other cards 😄

mattanglin: Cool 👍🏼 . I was using the typography variants listed in the figma file, but they did seem a little big. Will change these up.

yvonnetangsu: We do these push and pull with our design team often 😄 . Since sometimes things need adjustment once they've been built. Please feel free to do the same if you feel like designs could be tweaked.

I'm actually asking Bhavika more questions about the date block since the font size and alignment (left margin) are different from the event cards, but this PR doesn't have to wait for that, since we will have plenty of time to do those finetuning later on. I'll come back to review this some more after finishing up my accessibility PR for SAA so I'm not blocking you.

mattanglin: Cool. Maybe we can pull them out the resulting card styles into a more generalized component that can be used for all event/perk/trip cards

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
+export const root =
+  "su-group su-relative su-overflow-hidden su-bg-black-true su-break-words su-border-su-border-black";
+export const content = "su-px-26 su-pb-34 su-text-white";
```
yvonnetangsu, line 3: For the padding and margins, we prefer to use the responsive spacing classes when we can, e.g., su-rs-px-1, su-rs-mb-0 etc. You could probably use the same values for the paddings as the Event cards for this one.

src/components/cards/TripCard/TripCard.styles.js
yvonnetangsu: Yeah, I got the "GradientCard" one for Perk/Collection cards, and I can see another abstract style for the Event/Trip cards (we have so many cards 😂 ). I can see both of us doing some housecleaning this upcoming sprint too. For this PR, totally fine with how this is for now, so this can be merged sooner.

yvonnetangsu: ^^^ and of course the GradientCard could use some refactor - I like the new convention you set in this PR.

mattanglin: Yeah, I saw the `GradientCard` and started out trying to use it for trips then realized I was about to make a mess of the existing code and decided to do a new one instead 🤷🏼

yvonnetangsu: The 2 styles are different enough so it's probably easier to have 2 separate ones anyway 😄  Reusing only make sense when it doesn't cause a lot of hair pulling.

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
+export const root =
+  "su-group su-relative su-overflow-hidden su-bg-black-true su-break-words su-border-su-border-black";
+export const content = "su-rs-px-1 su-rs-pb-3 su-text-white";
+export const image =
+  "su-w-full su-h-full su-transition-transform su-transform-gpu group-hover:su-scale-[1.03] group-focus-within:su-scale-[1.03]";
+export const date =
+  "su-relative su-mt-[-5.6rem] lg:su-mt-[-6.3rem] su-mb-26 su-z-10";
+export const link =
+  "su-stretched-link su-z-20 su-rs-mt-0 su-mb-08em su-no-underline hocus:su-underline su-underline-offset !su-underline-thick !su-underline-digital-red-xlight";
```
yvonnetangsu, line 9:
```suggestion
  "su-stretched-link su-z-20 su-rs-mt-0 su-mb-08em su-type-3 su-no-underline hocus:su-underline su-underline-offset !su-underline-thick !su-underline-digital-red-xlight";
```
Give this a try, in combination with the below su-type-0 in heading

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
...
+export const heading =
+  "su-mt-32 su-type-3 su-font-bold su-relative su-inline su-text-white";
```
yvonnetangsu, line 11:
```suggestion
  "su-mt-32 su-type-0 su-font-bold su-relative su-inline su-text-white";
```

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
...
+export const icon = "su-relative su-inline-block";
```
yvonnetangsu, line 12:
```suggestion
export const icon = "su-relative su-inline-block group-hocus:su-text-white";
```
For headlines on dark bg, icon changes to white on hover/focus

yvonnetangsu:
@mattanglin If you could please add one schema called headingLevel (dataSource Heading Level available) for the trip card, that'd be great. We add that for all the cards (please see event cards for example), so for accessibility, content authors can pick which heading level depending on how deeply nested these cards are.

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
...
+  'su-group su-relative su-overflow-hidden su-bg-black-true su-break-words su-border-su-border-black';
+export const content = 'su-rs-px-1 su-rs-pb-3 su-text-white';
+export const image =
+  'su-w-full su-h-full su-transition-transform su-transform-gpu group-hover:su-scale-[1.03] group-focus-within:su-scale-[1.03]';
+export const date =
+  'su-relative su-mt-[-5.6rem] lg:su-mt-[-6.3rem] su-mb-26 su-z-10';
+export const link =
+  'su-stretched-link su-z-20 su-rs-mt-0 su-mb-08em su-type-3 su-no-underline hocus:su-underline su-underline-offset !su-underline-thick !su-underline-digital-red-xlight';
+export const heading =
+  'su-mt-32 su-type-0 su-font-bold su-relative su-inline su-text-white';
```
mattanglin, line 11: hmmmm... I think these style classes are still off. This is resulting in a trip card heading computed `font-size` of `44.85px` while the event card heading computes to `28.75px`...

yvonnetangsu: That's because event cards have 2 heading size options, the large one uses su-type-3 while the default one uses su-type-1 😄

mattanglin: Ah. I see now.

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
+export const root =
+  'su-group su-relative su-overflow-hidden su-bg-black-true su-break-words su-border-su-border-black';
```
yvonnetangsu, line 2:
```suggestion
  'su-group su-relative su-overflow-hidden su-bg-black-true su-break-words su-border-black';
```
Looks like a typo perhaps?

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
...
+export const content = 'su-rs-px-1 su-rs-pb-3 su-text-white';
```
yvonnetangsu, line 3:
```suggestion
export const content = 'su-rs-px-2 su-rs-pb-3 su-text-white';
```
For now, let's use a padding-2 so the content padding matches the event. On the event cards, the dateblock is actually pulled more to the left compared to the content, so I'm asking Bhavika to confirm this. This is her most updated mockup
Screen Shot 2021-09-14 at 4 45 22 PM

It's still different from events, but for now let's just do a su-rs-px-2 on everything.

yvonnetangsu: Also, if you could please make the month and day left aligned in the dateblock, that'd be great (this part we're sure about)

src/components/cards/TripCard/TripCard.styles.js
```diff
@@ -0,0 +1,14 @@
...
+  'su-w-full su-h-full su-transition-transform su-transform-gpu group-hover:su-scale-[1.03] group-focus-within:su-scale-[1.03]';
+export const date =
+  'su-relative su-mt-[-5.6rem] lg:su-mt-[-6.3rem] su-mb-26 su-z-10';
+export const link =
+  'su-stretched-link su-z-20 su-rs-mt-0 su-mb-08em su-type-3 su-no-underline hocus:su-underline su-underline-offset !su-underline-thick !su-underline-digital-red-xlight';
+export const heading =
+  'su-mt-32 su-type-0 su-font-bold su-relative su-inline su-text-white';
+export const icon = 'su-relative su-inline-block group-hocus:su-text-white';
+export const subheading = 'su-mt-4 su-type-0 su-font-semibold';
+export const description = 'su-mt-19 su-card-paragraph';
```
yvonnetangsu, line 14:
```suggestion
export const subheading = 'su-mt-4 su-mb-0 su-type-0 su-font-semibold';
export const description = 'su-rs-mt-0 su-mb-0 su-card-paragraph ';
```

yvonnetangsu (changes requested):
Looking awesome, @mattanglin 👏🏼  Just those few final things in suggestions, and one more:

Looks like somehow the dateblock has a z index higher than the stretched link itself so when you hover over it, it's not clickable. I think you're working on it already, but it might need some small tweaks still. Let's make it work like the event cards:
https://alumni-preview.stanford.edu/test/yvonne/test-events/

If you could please fix those, then this is GTG! We can finetune as we get confirmation from designer.

yvonnetangsu (approved):
Thank you for the changes @mattanglin , GTG 🚀  - we'll come back to those dateblock changes later. I like the new conventions and will try to match what you have from now on 😄

PR #182, opened 2021-09-22
ADAPT-3286 | @mattanglin | trip filter chips

mattanglin:
# READY FOR REVIEW

# Summary
- Styled trip filter chips

# Review Tasks

1. Navigate to https://deploy-preview-182--stanford-alumni-uat.netlify.app/travel-study/destinations/?page=1&trip-experience=expeditions-safaris&trip-experience=in-depth-study&trip-month=jun&trip-region=africa&trip-year=2022
3. Verify filter chips look and behave as expected


src/tailwind/plugins/theme/colors.js
```diff
@@ -30,6 +30,9 @@ module.exports = function () {
       DEFAULT: '#505EEC',
       light: '#6B77F5',
     },
+    'saa-electric-purple': {
+      DEFAULT: '#505EEC',
+    },
```
yvonnetangsu, line 35: We have this color already - line 30 🤣

yvonnetangsu (commented):
Works great! Couple really minor suggestions

- Please add a 0.3em left margin `su-ml-03em` on the X icon in the chips
- Please add a -0.7rem top margin on the Clear all button wrapper to lift it up a bit so it visually aligns a bit more centered with the chips. The chips have margin on the bottom so that messes with the flex alignment a bit 😄 

After that GTG - thanks for all the aria-labels too!

src/tailwind/plugins/theme/colors.js
mattanglin: > We have this color already - line 30 🤣

🤦🏼  We sure do. I just saw the color name in figma and assumed they were different 😝

yvonnetangsu:
@mattanglin Just found one more - for both the chips, and the filter checkboxes, the gradient direction should be `bottom left` instead of `top right` 😄
Screen Shot 2021-09-23 at 10 43 21 AM

mattanglin:
> @mattanglin Just found one more - for both the chips, and the filter checkboxes, the gradient direction should be `bottom left` instead of `top right` 😄
> Screen Shot 2021-09-23 at 10 43 21 AM

good catch! updated 👍🏼

yvonnetangsu (approved):
Nice - GTG!

PR #192, opened 2021-09-28
merge main back into dev

mattanglin:
# NOT READY FOR REVIEW
Do not merge this PR. It is merely a placeholder to have conversations about merge conflicts.


tailwind.config.js
```diff
@@ -32,7 +29,7 @@ module.exports = {
     lineHeight: require(decanter + "/theme/lineHeight.js")(),
     maxHeight: require(decanter + "/theme/maxHeight.js")(),
     maxWidth: require(decanter + "/theme/maxWidth.js")(),
-    screens: require(decanter + "/theme/screens.js")(),
+    screens: require(dir + "/theme/screens.js")(),
```
mattanglin, line 32: @yvonnetangsu I'm getting a conflict here that I'm not sure how to resolve. It looks like the screens config dir was changed to local, but also added to the extend section. What should the config for screens actually resolve to for both theme and extend sections?

yvonnetangsu: Yup, please keep the theme config dir to local. The extend for screen can be removed.

mattanglin: > Yup, please keep the theme config dir to local. The extend for screen can be removed.

Thank you! 🎉

PR #202, opened 2021-10-01
ADAPT-3817: replaced react-loader-spinner package with react-spinners…

andrewbaraniuk:
… package

# READY FOR REVIEW

# Summary
- Uses react-spinners/ClipLoader spinner instead of the old one..

# Review By (Date)

# Criticality

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "react-loader-spinner": "^3.1.14",
     "react-player": "^2.9.0",
     "react-refresh": "^0.8.3",
+    "react-spinners": "^0.11.0",
     "sanitize-html": "^2.4.0",
     "sharp": "^0.28.3",
     "storyblok-react": "^0.1.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,7 @@
         "react-loader-spinner": "^3.1.14",
         "react-player": "^2.9.0",
         "react-refresh": "^0.8.3",
+        "react-spinners": "^0.11.0",
         "sanitize-html": "^2.4.0",
         "sharp": "^0.28.3",
         "storyblok-react": "^0.1.2",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "react-loader-spinner": "^3.1.14",
```
sherakama, line 45: Please remove this package.

yvonnetangsu (changes requested):
Tested locally - works well for me. GTG after removing old react-loader-spinner package as @sherakama mentioned.

package.json
andrewbaraniuk: Done!

yvonnetangsu (approved):
GTG - thanks, @andrewbaraniuk! 
@sherakama, are we ready to merge changes into `main`? 😂

yvonnetangsu:
@sherakama I merged `dev` into this branch after you changed the base to `dev` in order to resolve the merge conflicts, but maybe I should've waited because `dev` was behind `main` (didn't have the launch stuff in yet) 😬

sherakama:
Well, since this branch has everything now I decided to add the updates from Matt's starter package and begin prepping for the removal of the alumni-dev2.stanford.edu environment and switching to a branch build with the appropriate contexts and context variables. 

@yvonnetangsu and @mattanglin please to review.

yvonnetangsu:
@sherakama The preview build is having that weird issue with the main nav links again (the default/). It could be that I butchered the package-lock.json file 😬

sherakama:
@yvonnetangsu @mattanglin 

I think I tracked down the `/default` link issue. @mattanglin was on the nose when he identified it was something to do with languages. I did some digging on our packages and the update to storyblok-js-client and gatsby-source-storyblok ended up changing the version of the API they were using from `v1` to `v2`. V2 has some more options on how you want to automatically resolve link paths when using the `multilink` field. Now, you have to specify the type of resolution you want as it will try to resolve the path automatically for you based on language. For some reason, the URLs were only getting prefixed when pulling `draft` content. Specifying the `url` option for resolveLinks and using the resolved object in sbLink did the trick for me.

One more issue with the Alert component links from Decanter after the upgrade to fix.

src/utilities/sbLink.js
```diff
@@ -15,7 +14,8 @@ import HeroIcon from '../components/simple/heroIcon';
 const SbLink = React.forwardRef((props, ref) => {
   // Storyblok link object either has a url (external links)
   // or cached_url (internal or asset links)
-  let linkUrl = props.link?.url || props.link?.cached_url || '';
+  let linkUrl =
```
sherakama, line 17: Here, use the `story` object for link URL resolution.

sherakama:
mmmm. Perhaps I did not get them all. Seems to pop up in a few places still.

yvonnetangsu:
> @yvonnetangsu @mattanglin
> 
> I think I tracked down the `/default` link issue. @mattanglin was on the nose when he identified it was something to do with languages. I did some digging on our packages and the update to storyblok-js-client and gatsby-source-storyblok ended up changing the version of the API they were using from `v1` to `v2`. V2 has some more options on how you want to automatically resolve link paths when using the `multilink` field. Now, you have to specify the type of resolution you want as it will try to resolve the path automatically for you based on language. For some reason, the URLs were only getting prefixed when pulling `draft` content. Specifying the `url` option for resolveLinks and using the resolved object in sbLink did the trick for me.
> 
> One more issue with the Alert component links from Decanter after the upgrade to fix.

Interesting! Thank you for tracking it down, @sherakama 😄

yvonnetangsu:
> mmmm. Perhaps I did not get them all. Seems to pop up in a few places still.

That is strange.....since all links in the main nav use the SbLink so theoretically if it works for one, it should work for all of them 🤔

sherakama:
Hmmm. Looks like `full_slug` is also prefixed...

sherakama:
https://forum.storyblok.com/t/resolve-links-does-not-work-as-described-according-the-documentation/640

sherakama:
Ok, i've got to put a pin in this one and move on to something else for a bit. I cannot figure out for the life of me why some of the URLs are resolved, why some include the story object along with the link object, and why some are unresolved and do not include the story object in the link object. 

I am tempted to do a string replace at this point...

yvonnetangsu:
> Ok, i've got to put a pin in this one and move on to something else for a bit. I cannot figure out for the life of me why some of the URLs are resolved, why some include the story object along with the link object, and why some are unresolved and do not include the story object in the link object.
> 
> I am tempted to do a string replace at this point...

Thank you for looking into this @sherakama. I could pull what Andrew did into a fresh PR that's based from dev in the meantime, then we can come back to this later perhaps?

sherakama:
Thread posted to Discord: https://discord.com/channels/700316478792138842/700326489731891201/895026080648400977

yvonnetangsu:
> > > @sherakama @yvonnetangsu I decided to dig into this a little bit this morning and think I've figured out the problem: Looking at the package-lock differences between this branch and `dev` I was able to determine that somewhere in these changes the [storyblok-js-client](https://github.com/storyblok/storyblok-js-client) was upgraded from `v4.1.3` to `v4.1.5`. Likely due to their poor documentation, it seems we may have been consuming Storyblok links incorrectly and looking through the release history for the storyblok-js-client you can see a change in [v4.1.3](https://github.com/storyblok/storyblok-js-client/releases/tag/v4.1.4) that [altered the way links are resolved to include language](https://github.com/storyblok/storyblok-js-client/commit/8b91deb199951c2b2804e29739caffa6c287ed61) to support i18n and i10n. Apparently they didn't deem this as breaking change. If you don't document usage there's nothing to break right?!?! 😐
> > > tl;dr We can resolve the link issue presently in one of 2 ways:
> > > 
> > > * Add `storyblok-js-client@4.1.3` to our package to lock link resolution behavior the way we currently have it implemented
> > > * Adjust our `SbLink` component to render links based on the changes to the updated sb client (by either stripping away the language prefix or resolving url and following the nested story e.g. `linkUrl = link?.story?.full_slug;`
> > > 
> > > Regardless of implementation, we need to change resolveLinks back to 'url' since story resolution is eating up too much memory in our builds...
> > 
> > 
> > Yikes. Thank you for looking into this, @mattanglin ! I'll let you and @sherakama decide on the action plan.
> 
> @yvonnetangsu @sherakama I opted to just lock the storyblok-js-client version for the time being and left a note about revisiting it. With it locked in our package we shouldn't have any random storyblok api changes moving forward unless we make the change ourselves.

Thank you, @mattanglin! Sounds great to me!

mattanglin:
> > One of the other "fun facts" about this is that it is still possible to fetch from the API the URLs without the language prefix.
> > https://share.getcloudapp.com/rRuWymR0
> 
> Haha! Awesome. It must be the storyblok-js-client injecting the language prefix under the hood... 😐

In case y'all are curious I dug a little more and found this in the [API spec](https://www.storyblok.com/docs/api/content-delivery/v2#core-resources/stories/retrieve-one-story) (which is now getting set to 'default` in the storyblok-js-client in the recent update and can't be "un-set" in any way. Might be worth creating an issue in their repo.
image

PR #233, opened 2021-11-02
ADAPT-3983: Remove 403/404 paths as they are now redirects.

sherakama:
# READY FOR REVIEW

# Summary
- Fixes conflict with 404 & redirects

# Review By (Date)
- End of sprint

# Criticality
- medium

# Review Tasks

1. Review build logs for duplicate path errors (same 404 for page and redirect)
2. Try to navigate to the /404 page. Should be redirected to a friendly 404 page
3. Try to navigate to /purple-monkey/dishwasher and be redirected to PGNet

PR #242, opened 2021-11-10
ADAPT-3983: Remove 403/404 paths as they are now redirects. (#233)

sherakama:
# READY FOR REVIEW

# Summary
- Cherry picks commit 412c340ef0bce9884baa3bb8d595552b5b4b413f out of dev for deployment

# Review By (Date)
- Today would be nice

# Criticality
- Medium / High

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "saa-alumni",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "license": "MIT",
       "dependencies": {
         "@fortawesome/fontawesome-pro": "^5.15.4",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "saa-alumni",
-      "version": "0.1.1",
+      "version": "0.1.2",
```
yvonnetangsu, line 9: @sherakama Is this behind?

yvonnetangsu (approved):
Thanks for this! The changes look good to me so preapproving. Got a question about the version though 🤔

package-lock.json
sherakama: hmm perhaps.

sherakama: Ah, that's just package-lock. This is fine.

yvonnetangsu: Cool - all GTG then! 🎉

mattanglin (approved):
lgtm 👍🏼 . I assume the redirects are being handled at our DNS level now?

sherakama:
> lgtm 👍🏼 . I assume the redirects are being handled at our DNS level now?

It is a mix. We have redirects from storyblok and custom 404/403 pages set up there now that this code is overriding.

PR #247, opened 2021-11-16
nojira | @mattanglin | these alerts are just wrong

mattanglin:
# READY FOR REVIEW

# Summary
Fix alert component to be hidden by default until locally stored dismissal checked to prevent server rendering

# Review By (Date)
- ASP
- 
# Review Tasks

1. Check that alerts only display on client (not in page source) when not already dismissed

### DIFF
--- a/src/components/composite/alert.js
+++ b/src/components/composite/alert.js
@@ -10,7 +10,7 @@ const Alert = ({
   blok,
 }) => {
   const hasCta = getNumBloks(alertCta) > 0;
-  const [isAlertDismissed, setIsAlertDismissed] = useState(true);
+  const [showAlert, setShowAlert] = useState(false);
   const isLinkDark = type === 'warning';
   let footerContent = '';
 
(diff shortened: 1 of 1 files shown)


src/components/composite/alert.js
```diff
@@ -10,7 +10,7 @@ const Alert = ({
   blok,
 }) => {
   const hasCta = getNumBloks(alertCta) > 0;
-  const [isAlertDismissed, setIsAlertDismissed] = useState(true);
```
yvonnetangsu, line 13: Oh interesting...I didn't even realize that the default was set to true 😮

mattanglin: The old variable names were terrible and the actual behavior was the opposite of what was actually happening

yvonnetangsu: That would explain the weird behavior. I was only looking at global alerts and didn't even look into this base component 🤦🏼‍♀️  Thanks for the fix

yvonnetangsu (approved):
Thanks for this!!

src/components/composite/alert.js
mattanglin: Hope you don't mind the PR. There's a lot of little things I've seen in the codebase that I've wanted to refactor because they seemed a little off and this was just one of them

yvonnetangsu: I don't mind at all - this is awesome! 😄

PR #264, opened 2021-12-16
ADAPT-3993: Filter out the past trips

Lyafel:
# Summary
- Filtered out the past trips to make sure none appear if the trip was not manually removed.

# Review By (Date)

# Criticality

# Review Tasks

1. Check out this branch
2. /travel-study/destinations/
3. Verify no past trips appear

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -54,9 +54,10 @@ exports.createPages = ({ graphql, actions }) => {
 
         const entries = result.data.allStoryblokEntry.edges;
         entries.forEach((entry, index) => {
+          let render = true;
           let slug = `${entry.node.full_slug}`;
           slug = slug.replace(/^\/|\/$/g, '');
-          let pagePath = entry.node.full_slug === 'home' ? '' : `${slug}/`;
+          const pagePath = entry.node.full_slug === 'home' ? '' : `${slug}/`;
 
           // Wire up the 404 page by setting the path to just 404 as Gatsby expects it.
           // if (pagePath.match(/^404/)) {
@@ -78,17 +79,30 @@ exports.createPages = ({ graphql, actions }) => {
             isCanonical = false;
           }
           const noIndex = content.noIndex ? content.noIndex : false;
+          // Don't create page for past trips.
+          if (content.component === 'trip') {
+            const start = new Date();
+            start.setHours(0, 0, 0, 0);
+            const now = Date.parse(start);
 
-          createPage({
-            path: `/${pagePath}`,
-            component: storyblokEntry,
-            context: {
-              slug: entry.node.full_slug,
-              story: entry.node,
-              isCanonical,
-              noIndex,
-            },
-          });
+            const tripDate = Date.parse(content.startDate);
+            if (now > tripDate) {
+              render = false;
+            }
+          }
+
+          if (render) {
+            createPage({
+              path: `/${pagePath}`,
+              component: storyblokEntry,
+              context: {
+                slug: entry.node.full_slug,
+                story: entry.node,
+                isCanonical,
+                noIndex,
+              },
+            });
+          }
         });
       })
     );
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -54,7 +54,6 @@ exports.createPages = ({ graphql, actions }) => {
 
         const entries = result.data.allStoryblokEntry.edges;
         entries.forEach((entry, index) => {
-          let render = true;
           let slug = `${entry.node.full_slug}`;
           slug = slug.replace(/^\/|\/$/g, '');
           const pagePath = entry.node.full_slug === 'home' ? '' : `${slug}/`;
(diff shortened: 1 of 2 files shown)


Lyafel:
For testing I unchecked the Hide on Trip Filter Pages from https://app.storyblok.com/#!/me/spaces/118608/stories/0/0/79657994 so it's supposed to appear in the list without the current pr.

sherakama:
@Lyafel Can you also add a condition to gatsby-node.js to filter out the past trips so that they don't get built at build time?

@yvonnetangsu to review.

yvonnetangsu (changes requested):
@Lyafel This looks good to me - please see @sherakama's comment about the gatsby node change, after that should be GTG. Thanks.

Lyafel:
@sherakama ready for review, hope I understood you correctly.

gatsby-node.js
```diff
@@ -78,17 +79,30 @@ exports.createPages = ({ graphql, actions }) => {
...
-            path: `/${pagePath}`,
-            component: storyblokEntry,
-            context: {
-              slug: entry.node.full_slug,
-              story: entry.node,
-              isCanonical,
-              noIndex,
-            },
-          });
+            const tripDate = Date.parse(content.startDate);
```
sherakama, line 88:
```suggestion
            const tripDate = Date.parse(content.endDate);
```

gatsby-node.js
```diff
@@ -78,17 +79,30 @@ exports.createPages = ({ graphql, actions }) => {
             isCanonical = false;
           }
           const noIndex = content.noIndex ? content.noIndex : false;
+          // Don't create page for past trips.
+          if (content.component === 'trip') {
+            const start = new Date();
```
sherakama, line 84: This might get a bit trixy. @yvonnetangsu the trip is assumed to start/stop on the PST timezone and I believe Netlify's servers run with UTC. With the timezone unset, it might unpublish the trip 8 hours too early. 

When creating the new date object you should set the timezone 

eg: `const start = new Date().toLocaleString("en-US", {timeZone: "America/Los_Angeles"})`

gatsby-node.js
sherakama: Check to see if this needs to be adjusted for timezone as well.

sherakama:
@yvonnetangsu I'd like to confirm my assumption of unpublishing trips that have passed. You do not want to display trips that have gone past their end date correct?

gatsby-node.js
yvonnetangsu: @sherakama For T/S we actually use the date picker that has no time info (only date), so we're actually using UTC for everything.

yvonnetangsu:
> @yvonnetangsu I'd like to confirm my assumption of unpublishing trips that have passed. You do not want to display trips that have gone past their end date correct?

I believe that's what they want. They're still encouraged to unpublish manually, but automatically removing it was a nice to have. Tagging @mattanglin to confirm since he was in some of the earlier discussions.

gatsby-node.js
sherakama: Aye, but the customer and client expectations would be that the start/stop for the trip be in PST isn't it? I understand the data from storyblok is one thing but there might be confusion if the trip goes away at 4:00PM on the day it ends, no?

mattanglin:
> > @yvonnetangsu I'd like to confirm my assumption of unpublishing trips that have passed. You do not want to display trips that have gone past their end date correct?
> 
> I believe that's what they want. They're still encouraged to unpublish manually, but automatically removing it was a nice to have. Tagging @mattanglin to confirm since he was in some of the earlier discussions.

@yvonnetangsu @sherakama From my recollection the idea was simply to remove the trips from showing up in the filters since they would would sorted to the top in the filtered result set, so just the work that's been done in `useTrips.js`. I'd say the changes in `gatsby-node.js` to not publish the trips is nice-ish to have, but also kind of moot since it requires the site to be rebuilt to affect anything 🤷🏼 . The general idea is that content authors would manually go in and unpublish past trips

gatsby-node.js
yvonnetangsu: @sherakama I see what you're saying, and my intuition totally agrees with you. But I've been tricked by this time thing before so I'm going to test this out a bit more myself 😂

sherakama: @Lyafel see previous comment.

yvonnetangsu: @sherakama I thought about this a little bit more, and I'm wondering that perhaps we could leave this as is. Reason being that when I'm looking at the trips on the site, it looks like they trip start date is actually the flight arrival date in the destination country 🤔  Here's an example of a trip to Japan - looks like the start date actually corresponding to the arrival date in Japan in the Itinerary section:
https://staging.alumni.stanford.edu/travel-study/destinations/japan-2022#itinerary-section

Plus with the datepicker without the time field, the date doesn't change no matter what the client's time zone setting is since we manually set it to UTC. In a previous PR, I had to remove the LA timezone part since it caused a discrepancy between the date shown in the Storyblok datepicker and the date actually rendered on the Trip page. So maybe if we just keep everything in UTC then it will be ok?
https://github.com/SU-SWS/saa_alumni/pull/223#discussion_r733967183

sherakama: Hmm. Well, it seems like my ask was not so simple after all. Without a clear timezone for the end date, I think we should just remove the gatsy-node.js code. I wouldn't want confusion between the build code and the runtime code. 

My apologies for overcomplicating this.

yvonnetangsu: @sherakama Sounds good to me 😄  Time is complicated - you're totally not overcomplicating it. Thank you for the discussion as usual. I appreciate you sharing your thoughts always!

@Lyafel Please remove the gatsby-node.js code as Shea suggested above.

sherakama (approved):
GTG

sherakama:
@yvonnetangsu gets to push the green button when ready.

yvonnetangsu:
> @yvonnetangsu gets to push the green button when ready.

GTF. Green button will be pushed 😄

yvonnetangsu (approved):

PR #277, opened 2022-01-25
ADAPT-4168: Implemented authentication workflow and protected page proof-of-concept

nplowman:
# READY FOR REVIEW
- DO NOT MERGE until redirect issue on Netlify is resolved.

# Summary
- Adds authentication, account links, and protected page proof-of-concept.

Includes the following tickets:
- ADAPT-4168
- ADAPT-4169
- ADAPT-4170
- ADAPT-4171
- ADAPT-4172

It also depends on unmerged work in the adapt-auth-sdk repo which can be found in https://github.com/SU-SWS/adapt-auth-sdk/pull/1. (The package.json file currently points at this feature branch directly to facilitate testing.) 

# Review By (Date)
- By end of current sprint. (Tuesday Feb, 2nd)

# Criticality

# Review Tasks

There is currently a strange redirect behavior affecting the Netlify environments. Until that is resolved, you can test this locally. 

1. You will first need some environment variables from Vault. Add the values for VAULT_ROLE_ID and VAULT_SECRET_ID to your .env file, and then run 'npm run vault'
2. Run 'netlify dev' to spin up the local environment.

### Typical Login Flow
1. Click the login link in the header.
2. Sign in with a set of valid credentials.
3. You should be redirected to the page you were on previously, and you should see the Account Links updated to show your username and the logout link in the dropdown.
4. Click "Logout" from the Account dropdown menu. Your session should be terminated, and you'll be redirected to the homepage.

### Protected page proof-of-concept.
1. While logged out, try to visit /protected-page. 
2. You should be redirected to the Login Page. Login with a set of valid credentials.
3. You should be redirected back to /protected-page. This time, you should see the page content. (Very stark white page with a header that says "Authenticated page Demo"

### Authentication Timeout
1. Log in at /api/auth/login
2. Let the browser tab sit idle for 15 minutes - you should see the timeout modal.
3. Closing or clicking the stay button should reset the time for an additional 15 minutes.
4. Clicking logout button, or allowing timeout to complete should terminate the session. 

If you don't want to wait the full 15 minutes, you can manually override the timeout setting in src/components/auth/AuthIdleTimeoutOverlay.js when testing locally. 

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources


src/components/identity/masthead.js
```diff
@@ -76,6 +77,7 @@ const Masthead = ({
                 className="first:su-ml-0"
                 hasExternalIcon
               />
+              <AccountLinks />
```
nplowman, line 80: Currently, I'm hard coding the Account links into the Utility Nav for both the regular site header and the Travel Study header. If preferable, a component could be created to represent the Account Links in Storyblok so you can manage where it is placed within the menu. 

As it is now, it will always show up as the last item in the nav.

src/components/navigation/accountLinks.js
```diff
@@ -0,0 +1,77 @@
...
+  );
+};
+
+const AccountLinks = (props) => {
+  const ref = createRef();
+  const [expanded, setExpanded] = useState(false);
+  const redirectUnauthenticated = false;
+  const { user, isAuthenticated } = useAuth(redirectUnauthenticated);
+  const loginDestination =
+    window !== 'undefined' ? window.location.pathname : null;
```
nplowman, line 21: Ideally, we'd be able to pull the current pathname serverside so this query parameter is set properly in both the pre-rendered markup as well as the hydrated page. 

It looks like Gatsby provides a location object at the page level that is available serverside, but we'd have to do pretty heavy refactoring to make sure it is passed down through several layers of components to use it here. 

Open to suggestions if there's a less invasive way to get at the pathname.

src/hooks/useAuth.js
```diff
@@ -0,0 +1,44 @@
+import { useState, useEffect } from 'react';
+import fetch from 'node-fetch';
```
sherakama, line 2: Need to look into this as this should be client-side fetch.

nplowman:
Also wanted to note that the test page (/protected-page) should probably be deleted from the code before merging this. 
It is helpful for validating the authentication while testing, but doesn't make sense to deploy to production.

package.json
```diff
@@ -15,8 +15,11 @@
     "@heroicons/react": "^1.0.4",
     "@tailwindcss/aspect-ratio": "^0.2.1",
     "@tailwindcss/forms": "^0.3.4",
+    "adapt-auth-sdk": "github:SU-SWS/adapt-auth-sdk#ADAPT-4167",
```
nplowman, line 18: Making a note that we should remove the #ADAPT-4167 reference after the work on adapt-auth-sdk is merged. (We should probably do that before merging this PR)

lambda/auth.js
```diff
@@ -0,0 +1,43 @@
...
+  '/api/auth/session',
+  authInstance.authorize({ allowUnauthorized: true }),
+  (req, res, next) => {
+    if (req.user) {
+      res.json(req.user);
+    } else {
+      res.status(401).json('UNAUTHORIZED');
+    }
+  }
+);
```
mattanglin, line 40: If we don't set `{ allowUnauthorized: true }` wouldn't we avoid having inspect the auth user?

src/components/auth/AuthIdleTimeoutOverlay.js
```diff
@@ -0,0 +1,81 @@
...
+  const handleLogout = () => {
+    window.location.replace('/api/auth/logout');
+  };
+
+  const modalTimer = useTimer({
+    timeout: modalTimeout,
+    onTimeout: handleLogout,
+  });
+
+  // eslint-disable-next-line no-undef
```
mattanglin, line 22: What's eslint complaining about here? I don't see any references to undefined variables?

src/hooks/useAuth.js
```diff
@@ -0,0 +1,48 @@
...
+import AuthContext from '../contexts/AuthContext';
+
+export const useAuth = (redirectUnauthorized) => {
+  const { isAuthenticated, isAuthenticating, user, setAuthState } =
+    useContext(AuthContext);
+
+  useEffect(() => {
+    let isMounted = true;
+    const url = `${window.location.protocol}//${window.location.host}/api/auth/session`;
+    fetch(url)
```
mattanglin, line 12: Prior to making this request for the session we should set `isAuthenticating` to `true`. Currently it doesn't look like that value is being set (or used) anywhere. We might need to dispatch actions to a reducer rather than simple state with [`useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer) to avoid issues with the`useState` callback scope.
Also, we should check the values of `isAuthenticated` and `isAuthenticating` to avoid making any unnecessary session requests.

src/hooks/useAuth.js
mattanglin: Yeah, `node-fetch` is strictly a server side fetch implementation. We can just use the browser fetch since this will only ever be run client-side.

lambda/auth.js
nplowman: @mattanglin 
I think the reason I had it implemented this way was because there are plans to have pages that will be will have personalized content for authenticated users, but still be accessible to anonymous users.  (There's a few wireframes and screenshots showing a use case in https://stanfordits.atlassian.net/browse/ADAPT-3987)

My concern was that the code in the auth_sdk looks like it will redirect the user when we check the session if 'allowUnauthorized' isn't set, which wouldn't be desirable for this use case. But if I'm misunderstanding how that code works and that's not the case, I can certainly refactor this. 

(Line of code I was concerned about is here: https://github.com/SU-SWS/adapt-auth-sdk/blob/main/src/AdaptAuth.ts#L223)

mattanglin: @nplowman As long as we don't set a config value for `unauthorizedRedirectUrl` or pass a `{ redirectUrl }` paramter to the `.authorize` call, it should default to responding with a `401 'UNAUTHORIZED'` 👍🏼 .

src/hooks/useAuth.js
nplowman: Import for node-fetch has been removed.

src/hooks/useAuth.js
nplowman: I'll start taking a look at refactoring with useReducer().

src/components/auth/AuthenticatedPage.js
```diff
@@ -0,0 +1,14 @@
...
+import { useAuth } from '../../hooks/useAuth';
+
+const AuthenticatedPage = ({ children, redirectUnauthorized = true }) => {
+  const { isAuthenticated } = useAuth(redirectUnauthorized);
+
+  if (isAuthenticated) {
+    return <>{children}</>;
+  }
+
+  return <div>Redirecting to login page...</div>;
```
mattanglin, line 11: I imagine we'd want to render a loader of some sort here while we fetch the auth status since _technically_ the user is only redirected if authentication fails and require authentication on this page.  That way the server render for these pages will just be the initial loader as well.

src/pages/protected-page.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import AuthenticatedPage from '../components/auth/AuthenticatedPage';
+
+const ProtectedPage = (props) => (
+  <AuthenticatedPage>
+    <h1>Authenticated Page Demo</h1>
+    <div>This page should only be visible to authenticated users.</div>
+  </AuthenticatedPage>
+);
```
mattanglin, line 9: Depending on how secure we want our authenticated-only page data to be, we'll likely have to create a lambda to proxy fetch requests for page content data since our auth is client side and the generated page content would be accessible from the bundled React JS.

mattanglin:
Overall, the auth integration seems to be working great locally! 🎉

src/pages/protected-page.js
nplowman: Yes - we'll also need to make sure the page data is included in the GraphQL queries in gatsby-node.js, or else the content could end up in page-data.json files even if it isn't rendered on the page. 

I already did some proof of concept research on how we could handle this for the Perks promo codes as part of https://stanfordits.atlassian.net/browse/ADAPT-3987 

It would be easiest if we can follow some patterns in Storyblok (e.g. place all protected content under a folder called "protected") so we can just exclude everything from GraphQL at once and not have to worry about updating it.

lambda/auth.js
```diff
@@ -0,0 +1,43 @@
...
+app.get('/api/auth/login', authInstance.initiate());
+app.get('/api/auth/logout', authInstance.destroySession());
+app.get(
+  '/api/auth/session',
+  authInstance.authorize({ allowUnauthorized: true }),
+  (req, res, next) => {
+    if (req.user) {
+      res.json(req.user);
+    } else {
+      res.status(401).json('UNAUTHORIZED');
```
sherakama, line 37:
```suggestion
      res.status(401).send('UNAUTHORIZED');
```

src/pages/protected-page.js
sherakama: At the moment there is less to worry about protecting information from the public as there is a need for providing personalization to the authenticated user.

lambda/auth.js
nplowman: Implemented this change.

lambda/auth.js
nplowman: @mattanglin I tried removing the "allowUnauthorized" flag as suggested, but it seems to be redirecting to the /403-access-denied page with a 200 status, rather than returning a 401 status as expected. 

I left this alone for now, and just made the small tweak @sherakama suggested. 

I'm not totally sure what is triggering the redirect - I'm seeing it locally so it's probably a clientside redirect triggered by Gatsby vs. something in the Netlify redirects file or netlify.toml.

src/hooks/useAuth.js
nplowman: @mattanglin Just finished refactoring everything to use a reducer pattern. 
I ended up just consolidating everything into AuthContext, as the useAuth() hook felt extraneous after making this change.

src/components/auth/AuthenticatedPage.js
nplowman: I've added a text placeholder condition for the "isAuthenticating" state for now.
@sherakama Do we already have a loading spinner in one of the ADAPT sites I can borrow for this? Or will we need some design input?

src/pages/protected-page.js
nplowman: @sherakama On that topic... Did you want me to create a working example of the personalization use case? I didn't see a subtask for that in the Jira ticket, but seems like it would make sense to implement.

lambda/auth.js
mattanglin: @nplowman I believe that redirect is happening because of the [session configuration on the auth instance](https://github.com/SU-SWS/saa_alumni/pull/277/files#diff-9a8b04c44d9e0c75f6d1b1e62f36d7164e4062c104d18773ae9b97bf747e42c6R20). The `authorize` middleware first checks for a `redirectUrl` passed to the method call, then checks on the instance config. If an unauthorized redirect url is found it redirects, otherwise it returns `401` ([redirect unauthorized code](https://github.com/SU-SWS/adapt-auth-sdk/blob/main/src/AdaptAuth.ts#L220-L227)).

src/pages/protected-page.js
sherakama: Thanks for the offer but I think we can take it from the integration.

src/components/auth/AuthenticatedPage.js
sherakama: I don't think we have a consistent spinner for the ADAPT projects. Feel free to pull something out of the `react-spinners` npm package for now.

nplowman: @sherakama I added the "PulseLoader" from react-spinners.
I just have applied some simple styling so it will show up in the top center of the content area. 
Easiest way to see it is probably to enable throttling in Chrome Dev tools, and then hit /protected-page while logged out.

lambda/auth.js
nplowman: @mattanglin Yep - it looks like having a value set for session.unauthorizedRedirectUrl in the instance settings was the culprit. Just pushed out a fix to remove that setting and the extraneous code in the session callback.

sherakama:
@nplowman, 

I am changing the base branch to go into the new `tsgg` branch as that will be our working branch for the authentication and Give Gab form integration. Sorry for the bait n switch.

sherakama:
Adding @rebeccahongsf to reviewers so you can get familiar with the code.

PR #301, opened 2022-02-25
ADAPT-4437 ADAPT-4520 | @rebeccahongsf | clone megaprofile api and refactor oauth gateway

rebeccahongsf:
# READY FOR REVIEW

# Summary
- clone megaprofile api
- refactor oauth gateway

# Review By (Date)
- By end of sprint Mochi

# Review Tasks

1. Check out this branch
2. Navigate to [`/api/auth/login`](https://idcs-8368be3faf0542efbdb27ae2b33d5d80.identity.oraclecloud.com/sso/v1/user/login)
3. Login with `zguan` then navigate to [`/api/auth/megaprofile-data`](https://deploy-preview-301--stanford-alumni.netlify.app/api/auth/megaprofile-data)
4. Confirm that contact information for `zguan` is being returned.
5. Navigate to [`/api/auth/test-404`](https://deploy-preview-301--stanford-alumni.netlify.app/api/auth/test-404)
6. Confirm that a `Not Found` error is returned
7. Navigate to [`/api/auth/test-405`](https://deploy-preview-301--stanford-alumni.netlify.app/api/auth/test-405)
8. Confirm that a `method not allowed` error is returned
9. Review code and see if anything looks weird or should be reworked.

- ADAPT-4437
- ADAPT-4520


src/utilities/ApiGatewayAuth.js
```diff
@@ -1,29 +1,28 @@
...
-    this.grantType = 'CLIENT_CREDENTIALS';
+    this.token = config.token || false;
+    this.grantType = 'client_credentials';
   }
 
   isAuthenticated = () => !!this.token;
 
   authenticate = async () => {
-    const params = {
+    const body = {
```
rebeccahongsf, line 17: [ADAPT-4520: Refactor oauth gateway work](https://stanfordits.atlassian.net/browse/ADAPT-4520) has been done here 😃 👍

src/utilities/MegaProfile.js
```diff
@@ -0,0 +1,68 @@
+import axios from 'axios';
+import { ApiGatewayAuth } from './ApiGatewayAuth';
+
+export class MegaProfile {
+  constructor(config = {}) {
+    // Set variables
+    const host = config.host || process.env.MEGAPROFILE_HOST;
+    const path = config.path || process.env.MEGAPROFILE_PATH;
+    const builtUrl = `${host}${path}`;
+    this.url = config.url || process.env.MEGAPROFILE_URL || builtUrl;
```
rebeccahongsf, line 10: The megaprofile URL has been set to in [Vault to a specific endpoint](https://vault.stanford.edu/ui/vault/secrets/secret/show/projects/adapt/tsgg) using Guan Min Chow's encodedSUID

I assume we'll want to rework this a bit so that we can adjust the path to use the user's encodedSUID within the fetch and store? 🤔

src/api/api-gateway-test.js
```diff
@@ -0,0 +1,16 @@
+import { ApiGatewayAuth } from '../utilities/ApiGatewayAuth';
```
sherakama, line 1: Probably should get rid of this endpoint now that we know it works.

src/api/megaprofile-test.js
```diff
@@ -0,0 +1,16 @@
+import { MegaProfile } from '../utilities/MegaProfile';
+
+// Get A User Data. Delete this prior to merging branch into TSGG
+// Do not enable for public use. This is for development/debugging purposes only.
+// Data can be viewed at /api/megaprofile-test
+// -----------------------------------------------------------------------------
+
+export default async function handler(req, res) {
```
sherakama, line 8: This endpoint should be protected from non-authenticated users.

src/api/megaprofile-test.js
```diff
@@ -0,0 +1,16 @@
...
+  try {
+    const mp = new MegaProfile();
+    const { data } = await mp.get();
```
sherakama, line 11: I think that the get method should take two parameters. 
1. The endpoint path after all of the prefix and version stuff
2. The encoded SUID, which should be pulled from the logged-in user (JWT in a cookie has this info)

```suggestion
    const { data } = await mp.get('/profiles/contact', user.encodedSUID);
```

With these two params, we can also hit other endpoints like 

```suggestion
    const { data } = await mp.get('/profiles/affiliations', user.encodedSUID);
```

src/api/megaprofile-test.js
```diff
@@ -0,0 +1,16 @@
...
+    res.status(200).json(data);
+  } catch (e) {
+    res.status(500).json(e.response.data);
```
sherakama, line 14: You'll likely want to handle the 404 & 405 responses from the API as well. Check what the API might return if something went wrong: https://anypoint.mulesoft.com/exchange/3a83e2b6-bdb0-438b-8e73-bfaf4f7596db/as-profile-studentactivities/minor/1.0/console/method/%234124/

src/utilities/ApiGatewayAuth.js
```diff
@@ -1,29 +1,28 @@
...
       client_id: this.clientId,
       client_secret: this.clientSecret,
       grant_type: this.grantType,
     };
-    const result = await axios.post(this.url, undefined, {
-      params,
+    const result = await axios.post(this.url, qs.stringify(body), {
+      headers: { 'Content-type': 'application/x-www-form-urlencoded' },
     });
```
sherakama, line 24: In case something goes wrong you'll also want to handle errors as `result` might not always be a token.

src/utilities/MegaProfile.js
sherakama: Yeah, you'll want to provide `https://api-dev.stanford.edu/advancementsystems/v1/constituents` as your host and have the encodedSUID and remaining endpoint path passed in as part of the method that is being called's parameters.

src/utilities/MegaProfile.js
```diff
@@ -0,0 +1,68 @@
...
+    this.auth = config.auth || new ApiGatewayAuth();
+
+    // Create Client
+    this.client = axios.create({
+      baseURL: this.url,
+      // Other request defaults? Headers?
+      params: {
+        token: this.authenticate(),
+      },
```
sherakama, line 19:
```suggestion
      params: {
        token: this.auth.token.access_token,
      },
```

Might as well authenticate and fetch that token prior to setting the param or remove this whole part and just call the authenticate function directly here at the end of the constructor.

src/utilities/MegaProfile.js
```diff
@@ -0,0 +1,68 @@
...
+    this.client.defaults.params.token = this.auth.token.access_token;
+  };
+
+  /**
+   * Auth handler
+   */
+  authenticate = async () => {
+    await this.auth.authenticate();
+    return this.setToken();
+  };
```
sherakama, line 34: Does this need to be two functions? Also this.setToken doesn't return anything so nor does the authenticate method.

src/utilities/MegaProfile.js
sherakama: Remove this as it `this.authenticate()` never returns anything and you are setting the object value directly later.

src/utilities/MegaProfile.js
sherakama:
```suggestion
  /**
   * Auth handler
   */
  authenticate = async () => {
    await this.auth.authenticate();
     this.client.defaults.params.token = this.auth.token.access_token;
  };
```

src/utilities/MegaProfile.js
```diff
@@ -0,0 +1,68 @@
...
+
+    // Create Client
+    this.client = axios.create({
+      baseURL: this.url,
+      // Other request defaults? Headers?
+      params: {
+        token: this.authenticate(),
+      },
+    });
+  }
```
sherakama, line 21:
```suggestion
    this.authenticate();
  }
```

src/utilities/MegaProfile.js
```diff
@@ -0,0 +1,68 @@
...
+  put = async (url, data, config = {}) =>
+    this.request({
+      ...config,
+      method: 'PUT',
+      url,
+      data,
+    });
+
+  delete = async (url, config = {}) =>
+    this.request({ ...config, method: 'DELETE', url });
```
sherakama, line 67: We can leave these in here but with any luck, we won't need them. It is looking like we will be read-only after all.

src/utilities/ApiExceptions.js
```diff
@@ -0,0 +1,28 @@
+export const exceptionMessages = {
```
rebeccahongsf, line 1: Pull this utility func over from adapt-my-account to handle errors. See [ApiException.ts](https://github.com/SU-SWS/adapt-my-account/blob/dev/src/utils/ApiException.ts)

src/utilities/ApiGatewayAuth.js
```diff
@@ -1,29 +1,39 @@
...
-    this.token = result.data;
-
-    return this.token;
+    await axios
+      .post(this.url, qs.stringify(body))
+      .then((response) => {
+        this.token = response.data;
+        return this.token;
+      })
+      .catch((error) => {
```
rebeccahongsf, line 28: Err. I added some conditional console.log()s for Axios error handling but is there a more elegant way to present these errors? 🤔

sherakama: Logging the error to the console is good but you should also throw the error so that the MegaProfile client aborts the call to the API as you are not authenticated to do so.

src/api/auth/megaprofile-data.js
```diff
@@ -0,0 +1,37 @@
...
+} from '../../utilities/ApiExceptions';
+
+const megaprofileHandler = async (req, res) => {
+  const mp = new MegaProfile();
+  try {
+    const { user } = req;
+    const { data: contact } = await mp.get(
+      `/${req.user.encodedSUID}/profiles/contact`
+    );
+    if (!contact) return NotFoundException(res);
```
sherakama, line 20:
```suggestion
```
I don't think you need to return this explicitly here. If the try/catch catches the error from axios you should be covered. Otherwise, you will need to do something like 

```
const { data: contact } = await mp.get(
  `/${req.user.encodedSUID}/profiles/contact` 
)
.catch((err) => {
    return ExceptionHandler(res, err);
});
```

src/api/auth/test-404.js
```diff
@@ -0,0 +1,28 @@
+// TODO: DELETE BEFORE MERGING. THIS IS SOLELY FOR TESTING 404 ERROR HANDLING
```
sherakama, line 1: This is fine to delete now.

src/api/auth/test-405.js
```diff
@@ -0,0 +1,28 @@
+// TODO: DELETE BEFORE MERGING. THIS IS SOLELY FOR TESTING 405 ERROR HANDLING
```
sherakama, line 1: This is fine to delete now.

src/utilities/ApiGatewayAuth.js
```diff
@@ -1,29 +1,42 @@
...
+          console.log('Error:', error.response.status, error.response.data);
+          Promise.reject(error);
+        }
+        if (error.request) {
+          console.log('Error:', error.request);
+          Promise.reject(error);
+        }
+        console.log('Error:', error.message);
+        Promise.reject(error);
+      });
```
sherakama, line 40:
```suggestion
      .catch((error) => {
        console.log('Error:', error.config);
        console.log('Error:', error.message);

        if (error.response) {
          console.log('Error:', error.response.status, error.response.data);
        }
        
        if (error.request) {
          console.log('Error:', error.request);
        }
        
        Promise.reject(error);
      });
```

Just a note on style, you don't need to reject multiple times if all you are doing is logging to the console.

src/api/auth/megaprofile-data.js
```diff
@@ -0,0 +1,33 @@
+// Get MP User Data and append to User
```
sherakama, line 1: Just one last note on this file name as it could be shorter and the URL `/api/auth/megaprofile-data` isn't super descriptive. How about `/api/auth/profile` or move this to `/api/user/profile`

sherakama (approved):
A couple of minor suggestions but otherwise good stuff. Marking as GTG so you can merge and keep moving forward.

PR #312, opened 2022-03-16
ADAPT-4539: Protected content 

Lyafel:
# Summary
- Protected wrapper component and promo code theming
- Currently the logic for showing the alumni vs not is missing so the only available blocks are the anonymous and the promo code. the 3rd option is commented out in ProtectedContentWrapper.jsx an can be tested locally. 
- Created a test item on /test/elena/misc

# Review By (Date)

# Criticality

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources


lambda/private-proxy/private-proxy.js
```diff
@@ -0,0 +1,18 @@
+const StoryblokClient = require('storyblok-js-client');
+const storyblok = new StoryblokClient({
+  accessToken: process.env.GATSBY_STORYBLOK_ACCESS_TOKEN,
```
nplowman, line 3: The "GATSBY_" prefix exposes the Gatsby token on the clientside, which we'll need to avoid moving forward in order to protect private content.

@sherakama This means that any tokens which have been previously used should be considered compromised. So at some point, we'll need to decommission all of the old tokens and generate new ones. We'll also need to search the codebase and see if we're actually using this environment variable anywhere on the clientside, and refactor it so it proxies calls through a serverless function. 

Since that will require a bit of coordination, probably makes sense to leave this "as is" for the scope of this ticket, and make one or two tickets for addressing this before we deploy it to production.

nplowman: @Lyafel This looks like we have this function duplicated both as a Netlify Function and a Gatsby Function. 
We can kill the the one in the lambda folder, and just keep the one in /src/api.

sherakama: GATSBY_STORYBLOK_ACCESS_TOKEN should be the public token and that should be ok to have exposed overall. For this work, I would expect a new environment variable with the appropriate access to be available to the app but hidden from the front end.

Did something change that I am not aware of?

nplowman: We had also discussed looking into the "Access Token Scopes" app, which allows you to restrict access to certain folders for each API key. The big fallback with that was it only worked by whitelist. So on the "public" token, you would have to explicitly list every single folder on the site except for the "protected" content folder. 

And you would have to update this whitelist every time you added a new folder, which was far from ideal.

src/components/layout/ProtectedContentWrapper.jsx
```diff
@@ -0,0 +1,47 @@
...
+const ProtectedContentWrapper = ({ blok }) => {
+  const protectedContent = blok.protectedContent.story.full_slug;
+  const protectedContentInactive =
+    blok.protectedContentInactive.story.full_slug;
+  const [authenticatedContent, setAuthenticatedContent] = useState(null);
+
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      fetch(
+        // Temporarily there is no way to know yet if the user is an alumni or not, waiting on that.
```
sherakama, line 16: For reference later: https://asjira.stanford.edu/jira/browse/ADGI-846

src/components/layout/ProtectedContentWrapper.jsx
```diff
@@ -0,0 +1,47 @@
...
+        .then((pageContent) => {
+          setAuthenticatedContent(pageContent.story);
+        });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <AuthContext.Consumer>
+      {(authState) => (
```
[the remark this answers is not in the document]
sherakama: Would it be possible to instead of looking for the hardcoded string that matches, instead, provide the string from Storyblok? Is it possible to create a dataset on Storyblok and add it to this component that allows for the content author to pick multiple grants from that list and to assume show for all authenticated users if none are selected?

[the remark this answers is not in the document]
sherakama: That would also give us a way to test this component while we wait for the value to show up in the API.

sherakama:
Build errors on Netlify 

```
10:29:41 PM: failed Building static HTML for pages - 7.712s
10:29:41 PM: error Building static HTML failed for path "/authenticated-section-test/"
10:29:41 PM: 
10:29:41 PM:    6 |
10:29:41 PM:    7 | const ProtectedContentWrapper = ({ blok }) => {
10:29:41 PM: >  8 |   const protectedContent = blok.protectedContent.story.full_slug;
10:29:41 PM:      |                                                        ^
10:29:41 PM:    9 |   const protectedContentInactive =
10:29:41 PM:   10 |     blok.protectedContentInactive.story.full_slug;
10:29:41 PM:   11 |   const [authenticatedContent, setAuthenticatedContent] = useState(null);
10:29:41 PM: 
10:29:41 PM:   WebpackError: TypeError: Cannot read properties of undefined (reading 'full_slug')
10:29:41 PM:   
```

src/components/layout/ProtectedContentWrapper.jsx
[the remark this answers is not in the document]
nplowman: Sorry - completely missed this comment earlier. 

We'd need to modify the field schema for the Protected Content component that I built in the initial Proof-of-Concept, but that should be doable. 

I think we would instead create a separate component (Maybe name it something like Protected Content Reference?) that would have two fields: A link field, and an Role/Affiliation field, where we would enter the string that we want to match on. 

And then in the ProtectedContentWrapper component, we would remove the two link fields and replace it with a multiple value Blok reference field.

nplowman (commented):
Just my two cents this this PR has been open for a while...
Most of the open comments on this PR are things I can address in https://stanfordits.atlassian.net/browse/ADAPT-4552.
So I think this can probably be approved and merged as-is.

yvonnetangsu:
> Just my two cents this this PR has been open for a while... Most of the open comments on this PR are things I can address in https://stanfordits.atlassian.net/browse/ADAPT-4552. So I think this can probably be approved and merged as-is.

Sounds good to me, @nplowman! Sounds like Shea had already had an initial review and had discussions with you before he left. I'll go ahead and approve so it's not a blocker for you 👍🏼

yvonnetangsu (approved):
GTG - thanks @Lyafel and @nplowman !

PR #327, opened 2022-04-11
ADAPT-4815 | @rebeccahongsf | rework mp api endpoint to fullgg

rebeccahongsf:
# READY FOR REVIEW

# Summary
- rework mp api endpoint to fullgg

# Review By (Date)
- end of sprint P

# Review Tasks

1. [Login](https://deploy-preview-327--stanford-alumni.netlify.app) as test user (e.g. `zguan`) 
2. Confirm that the user's name appears on the navigation bar
3. Navigate to [`api/auth/profile`](https://deploy-preview-327--stanford-alumni.netlify.app/api/auth/profile) and confirm that the gg profile along with the affiliations endpoint data exists 
4. Navigate to [`/travel-study/additional-payment-form`](https://deploy-preview-327--stanford-alumni.netlify.app/travel-study/additional-payment-form) to confirm the form fields are being correctly populated
5. Review code and see if anything could be improved 😎 

- ADAPT-4815


netlify.toml
```diff
@@ -2,6 +2,7 @@
 # ##############################################################################
 [build]
   publish = "public"
+  functions = "lambda"
```
sherakama, line 5: Why are you going back to Netlify functions? We should stick to Gatsby Functions.

rebeccahongsf:
Updates from 1:1 
- [ ] Remove lambda funcs
- [ ] Ref. Nathan's work in Giving and adjust Gatsby API to work with GG full profile endpoint.

src/api/auth/gg-profile.js
```diff
@@ -0,0 +1,78 @@
+import axios from 'axios';
```
rebeccahongsf, line 1: Assuming we will be swapping back to the individual endpoints later on, I created a separate page for the full GG profile information 🤔

src/contexts/AuthContext.js
```diff
@@ -32,20 +35,30 @@ class AuthContextProvider extends React.Component {
...
+        this.dispatch({ type: 'setUser', payload: body });
         this.dispatch({ type: 'setAuthenticating', payload: false });
       } else {
         this.dispatch({ type: 'setAuthenticated', payload: false });
-        this.dispatch({ type: 'setUserProfile', payload: null });
+        this.dispatch({ type: 'setUser', payload: null });
         this.dispatch({ type: 'setAuthenticating', payload: false });
       }
     });
+    const ggUrl = `${window.location.protocol}//${window.location.host}/api/auth/gg-profile`;
```
rebeccahongsf, line 51: Created a separate fetch so that we can have the user's firstName, lastName, encodedSUID, etc data in the event that the user does not have data populated within Mega Profile.

src/components/navigation/accountLinks.js
```diff
@@ -79,8 +79,8 @@ const AccountLinks = ({ mainLinkClasses }) => {
                   className={`su-inline-block su-mr-10 ${
                     showDesktop ? '' : 'su-sr-only'
                   }`}
-                >{`Hi, ${authState.userProfile.user.firstName} ${authState.userProfile.user.lastName}`}</span>
-                <Initial string={authState.userProfile.user.firstName} />
+                >{`Hi, ${authState.user?.firstName} ${authState.user?.lastName}`}</span>
```
rebeccahongsf, line 82: Updated the targeted data from userProfile (contains MP data) to user (comes from SAML(?)). This should cover the use case of when a user is able to login but has no data within MegaProfile.

package.json
```diff
@@ -15,7 +15,7 @@
     "@fortawesome/react-fontawesome": "^0.1.15",
     "@headlessui/react": "^1.3.0",
     "@heroicons/react": "^1.0.4",
-    "@netlify/plugin-gatsby": "^2.0.3",
+    "@netlify/plugin-gatsby": "^3.0.0",
```
sherakama, line 18: This diff looks ok to me. 

https://github.com/netlify/netlify-plugin-gatsby/compare/plugin-gatsby-v2.0.3...plugin-gatsby-v3.0.0#files_bucket

src/api/auth/gg-profile.js
sherakama: I think this is a good idea overall. A couple of pros/cons to this approach to consider

Pro:
1. Less refactoring later as you can hit the multiple endpoints from the microservices megaprofile API and structure the results to match what you are passing through for the fullgg endpoint

Con:
1. You only have 10 seconds to hit all of the megaprofile API endpoints to gather the data when it is behind one lambda function and could run into timeout issues. 🤞 you don't as the API is supposed to be performant enough for that but it could be something you have to deal with later.

src/api/auth/gg-profile.js
```diff
@@ -0,0 +1,78 @@
...
+  throw new Error('Response did not contain access token');
+};
+
+/**
+ * Get profile data from the MEGA PROFILE API
+ */
+
+const profileFetcher = async (profileID, token) => {
+  let response;
+  const endpoint = `${process.env.MEGAPROFILE_PROFILE_URL}/${profileID}/${process.env.MEGAPROFILE_PROFILE_TYPE}`;
```
sherakama, line 33: You won't have to make this configurable with the changes to the API structures. You can remove the profile type environment var.

```suggestion
  const endpoint = `${process.env.MEGAPROFILE_PROFILE_URL}/${profileID}/fullgg`;
```

src/components/navigation/accountLinks.js
sherakama: The perks work is expecting to interact with the userProfile key. Please restore that convention. 

More comments around this are forthcoming.

src/contexts/AuthContext.js
```diff
@@ -3,6 +3,7 @@ import AuthIdleTimeoutOverlay from '../components/auth/AuthIdleTimeoutOverlay';
 import setGiveGabVars from '../utilities/giveGabVars';
 
 const initialAuthState = {
+  user: null,
```
sherakama, line 6: The profile information should all be in one place and splitting them up isn't ideal. Please stick to the userProfile key.

src/contexts/AuthContext.js
```diff
@@ -14,6 +15,8 @@ function authReducer(state, action) {
       return { ...state, isAuthenticating: action.payload };
     case 'setAuthenticated':
       return { ...state, isAuthenticated: action.payload };
+    case 'setUser':
+      return { ...state, user: action.payload };
```
sherakama, line 19: You don't need to create a new setUser function as you can re-use the setUserProfile function and call it multiple times in your code.

src/contexts/AuthContext.js
```diff
@@ -14,6 +15,8 @@ function authReducer(state, action) {
...
     case 'setUserProfile':
       return { ...state, userProfile: action.payload };
```
sherakama, line 21: This function replaces the userProfile key with the payload values. If you need to update userProfile instead of fully replacing it you have two options. 

1. Modify and send the whole userProfile object each time you call setUserProfile
2. Create an updateProfile function which does a merge instead of a replace.

src/contexts/AuthContext.js
```diff
@@ -32,20 +35,30 @@ class AuthContextProvider extends React.Component {
   }
 
   componentDidMount() {
-    const url = `${window.location.protocol}//${window.location.host}/api/auth/profile`;
-    fetch(url).then(async (res) => {
+    const userUrl = `${window.location.protocol}//${window.location.host}/api/auth/session`;
```
sherakama, line 38: You'll still need to call `/api/auth/profile` as that function will need to collect from multiple sources.

src/contexts/AuthContext.js
sherakama: It probably makes sense to move all of this code (fetching from the fullgg endpoint) into the `api/auth/profile` endpoint. That way, all of the initial fetching for the user can happen there.

src/contexts/AuthContext.js
rebeccahongsf: The `setUserProfile` functionality is not new and exists on both the [`perks`](https://github.com/SU-SWS/saa_alumni/blob/perks/src/contexts/AuthContext.js#L16) and `tsgg` branch. Did we need to modify it?

sherakama: Modify it for your needs. Perks is going to get into the `dev` branch before `tsgg` and you can reconcile any differences after `perks` is merged.

rebeccahongsf: If we revert back to fetching from one source (e.g. `/api/auth/profile` or `/api/auth/gg-profile`) I do not believe we'll have to implement a updateProfile function since both APIs will include the user session data from `/api/auth/session`. 

However, if we wanted to provide a user an `Failed to fetch information from MegaProfile` error alert from the AuthContext, then that updateProfile function to merge the data might help 🤔   I suppose it's dependent on how we want notify the user of a MegaProfile fetch error. 

Hmm, let me think about this and test the updateProfile func out but any and all suggestions are welcomed in the meantime 😃

src/contexts/AuthContext.js
rebeccahongsf: If we're using the Full GG Profile located at `/api/auth/gg-profile` right now, did we still want to attempt to collect data from the MP with multiple endpoints in `/api/auth/profile`? Or shall we leave that fetch out for now and rework it back in once the MP for those endpoints are ready?

rebeccahongsf:
Pulled over some of Nathan's authContext and accountLinks work from Perks and adjusted it a bit 😄  All seems to be working as expected. Going to move forward with merging this in. I've linked ADAPT-4815 to our retro review ticket for future ref!

PR #366, opened 2022-06-21
ADAPT-4892: Added authentication pass-through for Storyblok Editor Preview

nplowman:
#  READY FOR REVIEW

# Summary
- Adds authentication pass-through for Storyblok editor, so forms and other authenticated pages can be previewed from within the Storyblok editor interface.

# How this Works
When viewing an authenticated page, the AuthContext makes calls to /api/auth/session and /api/auth/profile to validate that the user is authenticated and retrieve user profile info. This code change adds a middleware to both of these endpoints. If it detects that the endpoint is being called from the Storyblok editor path and it contains a valid access token in the URL, it will intercept the normal behavior of this endpoint and instead return mock data to simulate the user being logged in. 

# Review By (Date)

# Criticality

# Review Tasks

1. Check out this branch locally.
2. Add a new environment variable to your .env file called "STORYBLOK_EDITOR_TOKEN". This should be set to the value of the Preview token for the Storyblok dev space. (e.g. The same value you see in the access_key query param in the Storyblok editor preview)
3. Start up your local environment.
4. Try to edit a protected form in Storyblok, such as http://app.storyblok.com/#!/me/spaces/145938/stories/0/0/135068135/blok/dd15357d-47a4-4267-b6c0-85af8906d775?s=1 
5. Set the editor URL to your local environment. It should load the form instead of redirecting you to the login page.

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources


rebeccahongsf (approved):
Test it out and it LGTM 🚀 

cc: @sherakama

src/utilities/isStoryblokEditor.js
```diff
@@ -0,0 +1,15 @@
+export const isStoryblokEditor = (req) => {
+  if (req.headers.referer) {
+    const { referer } = req.headers;
+    const url = new URL(referer);
+    const accessKey = url.searchParams.get('access_key');
```
sherakama, line 5: So this access key is a working `preview` API key. 

We should be able to validate it against the Storyblok API by getting the space ID from `https://api.storyblok.com/v1/cdn/spaces/me?token={access_key}`

Then we should be able to compare that to `window.storyblok.space_id` to see if they match as I believe storyblok sets a number of window variables in their iframe. 

What about that instead of adding and having to manage the access tokens in multiple places?

src/utilities/isStoryblokEditor.js
```diff
@@ -0,0 +1,15 @@
...
+
+    if (
+      url.pathname === '/editor' &&
```
sherakama, line 8: Check if the referrer is `app.storyblok.com` here too?

src/utilities/isStoryblokEditor.js
sherakama: Alternatively, It might be possible to pass props down from [editor.js](https://github.com/SU-SWS/saa_alumni/blob/ADAPT-4892/src/pages/editor.js#L145) into the[ authenticated component](https://github.com/SU-SWS/saa_alumni/blob/ADAPT-4892/src/components/auth/AuthenticatedPage.js#L5) to 'pass through'

sherakama:
@ericbakenhus and @mattanglin can you give this a sniff test and chime in with thoughts and opinions as well. Whenever we open a new access door I'd like to have many eyes.

src/utilities/isStoryblokEditor.js
nplowman: @sherakama No, the referer URL will be for the app. (http://localhost:8000/editor when testing on a local environment).
The preview is iframe-embedded, and I don't think it is possible to check the URL of the parent frame serverside. 

However, Storyblok does append some additional query params in the preview editor that could potentially be used for some additional validation. Here's what the full referer URL looks like. (I stripped out the key and token param)

http://localhost:8000/editor?access_key=redacted&path=travel-study/italy-extension/registration&_storyblok=135068135&_storyblok_c=registrationFormPage&_st
oryblok_tk[space_id]=145938&_storyblok_tk[timestamp]=1656078289&_storyblok_tk[token]=redacted&_storyblok_version=&_storyblok_lang=default&_storyblok_rel
ease=0&_storyblok_rl=1656078435819'

src/utilities/isStoryblokEditor.js
nplowman: I think the first ID should work fine. This is being done serverside, so window.storyblok will not be available. But the space_id is still passed down in the query params.

sherakama:
ping @ericbakenhus & @mattanglin for comment

PR #407, opened 2022-07-21
ADAPT-5012 | @rebeccahongsf | fix up registration interstitial page style

rebeccahongsf:
# READY FOR REVIEW

# Summary
- update body text to plain textarea field
- add `su-subheading` to body textarea field
- update H3 to use `size={5}`
- center remove/add button with traveler name
- remove grid gap and add `margin-bottom: 18px` (excluding last child) to traveler cards
- update text above card list to use responsive spacing

# Review By (Date)
- before end of Sprint W

# Review Tasks

1. Navigate to [`/travel-study/destinations/italy-extension/register`](https://deploy-preview-407--stanford-alumni.netlify.app/travel-study/destinations/italy-extension/register)
2. Review styles and confirm that the request changes have been implemented
3. Review code

- ADAPT-5012


src/components/page-types/registrationFormPage/interstitialPage.js
```diff
@@ -187,14 +185,20 @@ const InterstitialPage = (props) => {
...
+                      align="left"
+                      font="serif"
+                      className="su-mb-0"
+                    >
                       Add existing connections and past travelers to your trip
                     </Heading>
                   </GridCell>
                   <GridCell xs={12} md={7} lg={8}>
-                    <p className="su-intro-text">
+                    <p className="su-subheading su-mb-0 su-rs-mt-1">
```
yvonnetangsu, line 201: My bad - I must have mistaken the previously rich text block to be this paragraph instead of the one above. This one has the original `su-intro-text` class instead of `su-subheading`. My apologies 🙇🏼‍♀️ 😹

src/components/page-types/registrationFormPage/interstitialPage.js
```diff
@@ -187,14 +185,20 @@ const InterstitialPage = (props) => {
                     )}
                   </GridCell>
                 </Grid>
-                <Grid xs={12}>
+                <Grid xs={12} className="su-rs-mb-5">
```
yvonnetangsu, line 188: ``
```suggestion
                <Grid gap xs={12} className="su-rs-mb-5">
```

Tip: We need the gap here so the width of each grid columns is the same as the ones with the grid below 😄 Screenshots showing the reason:
Screen Shot 2022-07-26 at 11 23 57 PM
Screen Shot 2022-07-26 at 11 24 08 PM

yvonnetangsu:
```
                    {body && (
                      <p className="su-subheading su-text-center">{body}</p>
                    )}
```
For line 184 (out of bounds here), if we could add a su-mb-0 for the paragraph that'd be awesome 😄

src/components/page-types/registrationFormPage/tripTravelerCard.js
```diff
@@ -77,39 +77,39 @@ const TripTravelerCard = ({ traveler }) => {
...
-              <div className="su-rounded-full su-p-2 su-bg-gradient-to-bl su-from-saa-electric-blue su-to-palo-verde-dark su-mr-02em">
-                <HeroIcon
-                  iconType="plus"
-                  className="su-text-saa-electric-blue su-rounded-full su-bg-saa-black"
-                />
-              </div>
-              <p className="su-basefont-23 su-mb-0">Add traveler</p>
-            </>
-          )}
+          <FlexBox direction="row" alignItems="center" justifyContent="start">
```
yvonnetangsu, line 80: A bit of defensive CSS here. If we could add a su-shrink-0 (or su-flex-shrink-0) class to this element, that would prevent the button text from wrapping to 2 lines when the name gets long. I imagine you don't want this 😄 
Screen Shot 2022-07-26 at 11 37 43 PM

yvonnetangsu:
Looking great - thank you for the changes. Left a few more ideas for you 😄

src/components/page-types/registrationFormPage/tripTravelerCard.js
yvonnetangsu: And while we're here, if we could add a tad more spacing between the icon and "Add traveler" that'd be awesome.

sherakama:
> casually including a font weight adjustment for custom journeys and notify me 😬

Lol what?

rebeccahongsf:
> > casually including a font weight adjustment for custom journeys and notify me 😬
> 
> Lol what?

For some reason my previous font weight fix for custom journeys and notify me was dropped (most likely during a merge conflict fix)

src/components/page-types/registrationFormPage/interstitialPage.js
```diff
@@ -183,18 +181,26 @@ const InterstitialPage = (props) => {
...
+                      align="left"
+                      font="serif"
+                      className="su-mb-0"
+                    >
                       Add existing connections and past travelers to your trip
                     </Heading>
                   </GridCell>
                   <GridCell xs={12} md={7} lg={8}>
-                    <p className="su-intro-text">
+                    <p className="intro-text su-mb-0 su-rs-mt-1">
```
yvonnetangsu, line 203: Where did the su- prefix go my friend 🤣 ?

yvonnetangsu:
We're super close. Here are my final comments @rebeccahongsf 😄 

1. In general, I try to avoid using `su-grid-gap` for `FlexBox` for now. Even though flex box supports the CSS `gap` property now, but Safari (esp iOS Safari) was pretty late with adopting it, so some versions of Safari v14 still doesn't support the `gap` property 😬 
https://caniuse.com/flexbox-gap
Screen Shot 2022-07-28 at 3 14 21 PM

Technically we only support the last 2 versions of a browser but I try to not cut it too close 😂 

2. Perhaps something to check with Bhavika, it seems like this might look a little better if either a) both heading and the "No travelers have been selected" are center aligned or b) both left aligned 🤔 This looks a little bit off to me but please feel free to push back (totally nitpicky me)
Screen Shot 2022-07-28 at 3 06 30 PM

4. Could the "Next" button use the existing component `SAALinkButton` similar to how we used it on the Trip page? Are we using the bright red outline variant instead of the solid (seems like the Figma mock uses the outline style)?
https://github.com/SU-SWS/saa_alumni/blob/81688a62e27f80d01df0426ada85be162232b5bb/src/components/page-types/TripPage/TripPageOverviewSection.jsx#L162

4. Are we still displaying the text "You have not added any additional travelers" when one only added themselves?
Screen Shot 2022-07-28 at 3 26 02 PM

5. Let's use `su-text-18` for the "Remove" button on the right side bar
6. I'd probably want to make sure the spacing above and below the grey separator is the same (might be more obvious when there are more than 2 names

Screen Shot 2022-07-28 at 3 30 11 PM

That should be all - thank you so much for the changes Rebecca Hong and for putting up with me @rebeccahongsf 🤣

rebeccahongsf:
> 

1. Ahh! That is great to know and duly noted for future usage. Thank you, Yvonne!! (ง’̀-‘́)ง 
2. Bhavika and I are going to discuss this further and make adjustments in a future PR
3. Since we need to use the Gatsby `<Link />` to pass the traveler information, I've left the component as is but updated the styling and icon type to use Hero Icons 😄 
4. This is a great question! Same as #2, Bhavika and I are going to discuss this further and make adjustments (if any) in a future PR.
5. Sounds good! The `remove` text for the traveler list has been updated 👍 
6. Great catch — I've updated the spacing and also added in the button margin-top spacing that was missing before.

I'm going to merge this in so Bhavika and I can review during out next QA theme session 🚀  Thank you again for all your amazing feedback on this, @yvonnetangsu 😸

PR #425, opened 2022-08-09
ADAPT-5195 | @rebeccahongsf | conditionally extend loader duration

rebeccahongsf:
# READY FOR REVIEW

# Summary
- extend loader duration based on event listener
# Review Tasks

1. Login as a test user
2. Navigate to [`travel-study/destinations/italy-extensions/register`](https://deploy-preview-425--stanford-alumni.netlify.app/travel-study/destinations/italy-extension/register)
3. Confirm that the loader displays then transitions to form with no empty state in between

- ADAPT-5195

### DIFF
--- a/src/components/embed/dynaScript.js
+++ b/src/components/embed/dynaScript.js
@@ -1,3 +1,4 @@
+import { dcnb } from 'cnbuilder';
 import React, { useRef, useEffect, useState } from 'react';
 import ClipLoader from 'react-spinners/ClipLoader';
 import GiveGabErrorMessage from './giveGabErrorMessage';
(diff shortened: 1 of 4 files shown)


src/components/embed/dynaScript.js
```diff
@@ -32,26 +37,31 @@ const DynaScript = ({ errorBlok, src, id, ...props }) => {
 
     return () => {
       mounted = false;
+      script.removeEventListener('widgetRenderEnd');
```
sherakama, line 40: @rebeccahongsf 
`removeEventListenter` has two parameters

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

> Given an event listener previously added by calling [addEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), you may eventually come to a point at which you need to remove it. Obviously, you need to specify the same type and listener parameters to removeEventListener()

You also need to pass it the listening function so you will have to refactor your addEventListener call and pull the anonymous function out and into a variable.

PR #465, opened 2022-09-19
ADAPT-5425 | @rebeccahongsf | fix email and phone number types logic

rebeccahongsf:
# READY FOR REVIEW

# Summary
- fix email and phone number types logic

# Review By (Date)
- asap

# Review Tasks

1. Checkout branch locally
2. Enable mockserver by adding `MEGAPROFILE_MOCK=true` in the `.env`
3. Navigate to `/travel-study/destinations/italy-test/register/form`
4. Continue onward to the registration form.
5. Verify that the prefilled email and phone number matches the preferred typed listed in their MP data at `/api/auth/profile`
6. Feel free to change the pref email and phone number types under `src/utilities/mocks/fullgg.js` to confirm that the proper email and phone numbers are being prefilled.
7. Review code 😎 

- ADAPT-5425


src/utilities/giveGabVars.js
```diff
@@ -189,6 +194,12 @@ export const findEmail = (emails) => {
...
     }
+    if (!email) {
+      email = findEmailType(emails, 'SAA Email');
+    }
+    if (!email) {
+      email = findEmailType(emails, 'GSB Email');
+    }
     if (!email) {
       email = findEmailType(emails, 'Other Email');
     }
```
yvonnetangsu, line 205: @rebeccahongsf I wonder, is there a reason that this couldn't be combined into one statement? eg

```
  if (Array.isArray(emails)) {
    email = findPreferredEmail(emails);
    if (!email) {
      email =
        findEmailType(emails, 'Home Email') ||
        findEmailType(emails, 'Business Email') ||
        findEmailType(emails, 'SAA Email') ||
        findEmailType(emails, 'GSB Email') ||
        findEmailType(emails, 'Other Email');
    }
  }
```
 I feel like you might have a reason for that so just want to check with you 😄

yvonnetangsu (approved):
Everything works as described so preapproving - just a question about whether we can combine some of these if statements (email, phone etc)

Also, this is not related to this PR but saw a missing key error in the console when loading the form page. I think this could be fixed by removing the fragment tags on line 19 and 25 in `TripTravelerListItem` (not sure if it breaks other things so please double check if you decide to update this in this PR also) 😄

src/utilities/giveGabVars.js
rebeccahongsf: AH! Great catch, @yvonnetangsu!!! No reason aside from following the preexisting pattern that we had in Give. I'll push up the rework version 😉

PR #486, opened 2022-10-13
EXRESUPP-48 | @rebeccahongsf | add conditional to gatsby node for unpublished trips

rebeccahongsf:
# READY FOR REVIEW

# Summary
- add conditional to gatsby node for unpublished trips

# Review By (Date)
- asap

# Review Tasks

1. Login as zguan
2. Navigate to:
- [`travel-study/destinations/japan-test`](https://deploy-preview-486--stanford-alumni.netlify.app/travel-study/destinations/japan-test)
- [`travel-study/destinations/japan-test/register`](https://deploy-preview-486--stanford-alumni.netlify.app/travel-study/destinations/japan-test/register)
3. Confirm that you are redirected to the cardinalalumni 404 page since the Trip page is not published
4. Review code

- EXRESUPP-48


gatsby-node.js
```diff
@@ -137,29 +137,31 @@ exports.createPages = ({ graphql, actions }) => {
...
-            component: storyblokEntry,
-            context: {
-              slug: `${registrationEntry.node.full_slug}/form`,
-              story: registrationEntry.node,
-              isCanonical,
-              noIndex,
-            },
-          });
+          // Create the GG form page(s) only if the trip is published
+          if (content.trip.name) {
```
rebeccahongsf, line 141: @sherakama thoughts on this as a check? 🤔

sherakama: That _should_ work, however, is there a status key that might be more clear to read in the code? also, if we get into workflows later it might make sense to check a status key instead of a content value.

rebeccahongsf: Hmm, I poked around in the `trip` and `registrationFormPage` data and couldn't find a status key for the published/unpublished status. The existing published related data is around the dates of when it was first published 🤔

sherakama: Yeah, I had a quick peek. Looks like we can filter on published status by the API key but we don't get that information as a key on the story object. Ok, let's go with what you have and if we nee to revise later we can.

sherakama (approved):
OK

PR #511, opened 2022-11-09
EXRESUPP-61 | @jdwjdwjdw | Find out where the registration and notify forms are slipping into search form

jdwjdwjdw:
# READY FOR REVIEW

# Summary
- Find out where the registration and notify forms are slipping into search form
- Remove user-guide and reserve form URL's from sitemap (this is not the cause of the registration and notify forms slipping into the search, but also needed per Rebecca)

# Review By (Date)

# Criticality

# Review Tasks

1. Go to https://deploy-preview-511--stanford-alumni.netlify.app/sitemap/sitemap-0.xml and confirm that the user-guide and reserve form URL's are not included.
2. Navigate to https://deploy-preview-511--stanford-alumni.netlify.app/search/?page=0&q=italy and confirm that the form results we want removed are no longer displayed.

## Front End Validation

### Code

### Code security

## General

- JIRA ticket(s)
- Other PRs

# Resources


gatsby-config.js
```diff
@@ -96,11 +96,14 @@ module.exports = {
           `/travel-study/global-components/**`,
           `/travel-study/destinations/**/register`,
           `/travel-study/destinations/**/register/form`,
+          `/travel-study/destinations/**/reserve`,
+          `/travel-study/destinations/**/reserve/form`,
```
sherakama, line 100: `/travel-study/**/form`

sherakama: page.pageContext.content.component !== 'form-page'

In the filterPagers function.

sherakama:
Filtering on the front end won't work and we should prevent the bad data from going into the index instead. Please continue this PR and add to the XML sitemap filter function in gatsy-config.js logic to remove the form content types.

gatsby-config.js
jdwjdwjdw: Thanks @sherakama! I updated the filterPages function to remove the form pages, which is reflected in the sitemap preview at https://deploy-preview-511--stanford-alumni.netlify.app/sitemap/sitemap-0.xml. It captured all of those without needing to add the specific form URL's, so I went ahead and removed those URL's.

Still working on preventing the bad data from going into the index for our search issue, though.

sherakama: > Still working on preventing the bad data from going into the index for our search issue, though.

Outside of adding a filter for the form pages, which you have done here, are you looking into anything else?

jdwjdwjdw: I was able to figure out how to filter the multiple queries but based on the [Aloglia documentation here](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-null-or-missing-attributes/#creating-a-tag)

> The Algolia engine doesn't support filtering on null value or missing attributes. In other words, in the preceding example, if you wanted to combine books with a specific rating and books that aren't yet rated in the same filtering statement, this would require some modification of the data.

We would not be able to filter on null value or missing attributes (e.g. filters: 'NOT title: ""' throws an error). Should we look into creating `_tags` mentioned in the docs? We could add tags and filter by [boolean values as mentioned here](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-boolean/)

sherakama: No, I don't think we need to go down that path. Removing the pages that require authentication from the sitemap and providing guard code in the react component for missing values should be enough to get us to where we want to be.

sherakama:
Add nofollow to links

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-nofollow

jdwjdwjdw:
> Add nofollow to links
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-nofollow

@sherakama I added `nofollow` where I could. However, I'm seeing that all of these `notify` and `register` pages also include CTA buttons added within storyblok near the bottom of the pages, which can't be hardcoded it seems? Should I just add `nofollow` as the default value for all CTA buttons?
Screen Shot 2022-11-22 at 4 53 50 PM

Screen Shot 2022-11-22 at 4 56 07 PM

sherakama:
@jdwjdwjdw Thanks for looking into this further. No, unfortunately, defaulting to no-follow is not an option for the storyblok editor. I think we have done our due diligence here and I'll double down and see about adding some protection to the crawler to prevent these empty records as well.

sherakama (approved):
GTG.Ty.

PR #530, opened 2022-11-30
ADAPTTS-249: Dummy date

sherakama:
# READY FOR REVIEW

# Summary
- If no data is available from MP, set the birthday to undefined

# Review By (Date)
- when possible

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

- ADAPTTS-249

### DIFF
--- a/src/components/page-types/registrationFormPage/interstitialPage.js
+++ b/src/components/page-types/registrationFormPage/interstitialPage.js
@@ -90,7 +90,7 @@ const InterstitialPage = (props) => {
         su_relation: relationship?.relationshipType,
         su_dob: relationship?.relatedContactBirthDate
           ? formatUsDate(relationship?.relatedContactBirthDate)
-          : '',
+          : undefined,
         su_reg: 'Related contact',
         su_email: undefined,
         su_phone: undefined,
--- a/src/components/page-types/registrationFormPage/interstitialPage.js
+++ b/src/components/page-types/registrationFormPage/interstitialPage.js
@@ -145,7 +145,7 @@ const InterstitialPage = (props) => {
       phoneNumberTypeList,
       primaryRegistrantPhoneNumberType
     ),
-    su_dob: userProfile?.birthDate ? formatUsDate(userProfile?.birthDate) : '',
+    su_dob: userProfile?.birthDate ? formatUsDate(userProfile?.birthDate) : '00/00/0000',
     su_relation: 'Guest',
     su_reg: 'Primary registrant',
   };
--- a/src/components/page-types/registrationFormPage/interstitialPage.js
+++ b/src/components/page-types/registrationFormPage/interstitialPage.js
@@ -147,7 +147,7 @@ const InterstitialPage = (props) => {
     ),
     su_dob: userProfile?.birthDate
       ? formatUsDate(userProfile?.birthDate)
-      : undefined,
+      : '00/00/0000',
     su_relation: 'Guest',
     su_reg: 'Primary registrant',
   };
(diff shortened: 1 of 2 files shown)


src/components/page-types/registrationFormPage/interstitialPage.js
```diff
@@ -145,7 +145,9 @@ const InterstitialPage = (props) => {
       phoneNumberTypeList,
       primaryRegistrantPhoneNumberType
     ),
-    su_dob: userProfile?.birthDate ? formatUsDate(userProfile?.birthDate) : '',
+    su_dob: userProfile?.birthDate
+      ? formatUsDate(userProfile?.birthDate)
+      : undefined,
```
rebeccahongsf, line 150: While I am still unclear as to why the user's birthDates did not prefill, I believe this is the solution we are looking for. In the event a birthDate is not provided (or unavailable because the MP was down), set the value as undefined so that the user can input their own birthday. 

Apologies for the improper syntax before — `''` will still prefill the GiveGab field whereas in the event there is invalid or no data, the user should be given the opportunity to fill out the field with their own information.

Thoughts about this implementation, @sherakama ? 😃

sherakama: From our documentation, the flow we want is:

>  for the primary registrant, we are not showing the DOB if they have one on file. In the case where there is no DOB, we will pass along dummy data

A couple of things I think are needed here for more robustness. The idea is that you can never pass an empty value for the primary user into the form. That is where the `00/00/0000` default came from. 

1. The formatUsDate function will need to be validated that it is returning a string as it is possible that the data going in is bad and not a proper date. Bad dates could be returning blank strings.
2. When no birthdate is available (null) handle that as well.
3. Check to see if the validation on the dateofbirth field in GG allows the dummy data of `00/00/0000`
4. If the dummy data does not work, do what you are doing here and allow the user to add it.

rebeccahongsf: Okay, I think my most iteration should fulfill the requirements with one caveat. Since `00/00/0000` is an invalid date, I've passed in the current date instead.

Let me know your thoughts and if additional adjustments are needed. Thank you @sherakama ! 😄

sherakama: 🤔 Hmmm. That might not work out for the best either as that might have issues with the Adult/Minor thing on trips. 

Can you follow up with Kristin on this flow? Let her know that we need to have a valid date.  

It might be time to think about the flow and allow the user to correct the invalid date instead of defaulting to something.

rebeccahongsf: Sounds good. I'll follow up with Kristin about the preferred user flow. Thanks Shea!

sherakama: The notes we got from Kristin was to always hide the DOB field and never show it. If we are planning on showing it now, we can avoid the dummy data.

src/utilities/transformDate.js
```diff
@@ -18,5 +18,6 @@ export const formatUsDate = (tripDate) => {
   const month = date.getMonth() + 1;
   const day = date.getDate() + 1;
   const year = date.getFullYear();
-  return `${month}/${day}/${year}`;
+  const newDate = `${month}/${day}/${year}`;
+  return newDate.toString();
```
sherakama, line 22: Why is the toString needed here?

src/components/page-types/registrationFormPage/interstitialPage.js
sherakama: Also, yes, you should protect against a bad date from the API by showing the input field as well.

rebeccahongsf: Thanks Shea! I moved the conditional check back into the formatUsDate func. I think the current state should now fulfill the requirements.

rebeccahongsf (approved):
GTG and approved per Shea's comment above 😄

PR #563, opened 2023-01-23
ADAPT-5539: Protect against content references in create blocks.

sherakama:
# READY FOR REVIEW

# Summary
- A content reference from somewhere was sneaking in as a story object
- Not sure where it came from
- Added extra try/catch on postscribe 

# Review By (Date)
- End of the week

# Criticality
- 5/10

# Review Tasks

1. Review code
2. Pull down to local and set your STORYBLOK_ACCESS_TOKEN to point to the production environment
3. Spin up a local build
4. Navigate to the production storyblok space and edit the home page
5. Ensure you can change a field or value without it WSODing
6. Review preview build to ensure home page still loads with ads

### DIFF
--- a/src/components/partials/seo.js
+++ b/src/components/partials/seo.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { Helmet } from 'react-helmet';
-import SbEditable from 'storyblok-react';
 import useSiteMetadata from '../../hooks/useSiteMetadata';
 import transformImage from '../../utilities/transformImage';
 import useSubsite from '../../hooks/useSubsite';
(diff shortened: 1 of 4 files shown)


yvonnetangsu:
@sherakama, the preview build and my local build is missing the masthead, and other components that are referencing a published version that aren't added just on that page, e.g., Story cards, event cards, global footer 🤔

sherakama:
> the preview build and my local build is missing the masthead

Oh boy, I need more coffee. How'd I miss that!

Ok, I'll have another pass at this.

yvonnetangsu:
> 

Haha no worries! You have a lot on your plate, @sherakama 😸

src/components/partials/seo.js
```diff
@@ -89,30 +88,28 @@ const Seo = ({
   const canonicalUrl = getCanonicalUrl(blok, siteUrl, location);
 
   return (
-    <SbEditable content={blok}>
```
sherakama, line 92: You can't really use the visual editor to edit meta tags now can we?

sherakama:
@yvonnetangsu Ready for review. This time there is a banner!

src/components/partials/seo.js
yvonnetangsu: That's true - no need for that 👍🏼  @sherakama Looks like a new package-lock.json is generated in this PR. Is it meant to be there?

yvonnetangsu (approved):
I was able to edit on local with no white page error and netlify preview looks great. @sherakama my only question is with the `package-lock.json` file difference. Other than that, this is GTG 👍🏼  Thank you for this fix 😄

sherakama:
> my only question is with the package-lock.json file difference

Yeah, I ran an `npm install` so there is a little bit of an update.

yvonnetangsu:
> > my only question is with the package-lock.json file difference
> 
> Yeah, I ran an `npm install` so there is a little bit of an update.

Sounds good - thank you @sherakama !! 🚀

PR #568, opened 2023-01-24
ADAPTSM-118 | @rebeccahongsf | update select ux and add digitalName

rebeccahongsf:
# READY FOR REVIEW

# Summary
- update select ux 
- update membership card styling
- add digitalName

# Review By (Date)
- when possible

# Review Tasks

1. Navigate to [`membership/register`](https://deploy-preview-568--stanford-alumni.netlify.app/membership/register)
2. Verify that you are able to select between the two type of registrants without having to unselect one before selecting the other
3. Navigate to [`membership/register/related-contacts`](https://deploy-preview-568--stanford-alumni.netlify.app/membership/register/related-contacts)
4. Verify that the new contact styling still appears as expected
5. Select a registrant and continue to the form
6. Confirm that the digital name now displays:
image
7. Review code (confirm that disable styling has been removed) 👾 

- ADAPTSM-118
- ADAPTSM-120

### DIFF
--- a/src/components/page-types/membershipFormPage/membershipCard.js
+++ b/src/components/page-types/membershipFormPage/membershipCard.js
@@ -11,7 +11,6 @@ const MembershipCard = ({
   initial,
   newContact = false,
   memberData,
-  disabled = false,
 }) => {
   const [state, dispatch] = useContext(FormContext);
   const { registrantsData } = state;
@@ -21,7 +20,7 @@ const MembershipCard = ({
 
   const addRelationship = () => {
     dispatch({
-      type: 'addRegistrant',
+      type: 'addSingleRegistrant',
       payload: memberData,
     });
   };
@@ -41,23 +40,13 @@ const MembershipCard = ({
     }
   };
 
-  let newContactWrapper = '';
-
-  if (newContact && isSelected) {
-    newContactWrapper = 'su-bg-saa-black';
-  } else if (newContact) {
-    newContactWrapper = 'su-border-dashed';
-  }
-
   return (
     <FlexBox direction="col" as="article" className={styles.root}>
       <button
         type="button"
         className={dcnb(
-          isSelected
-            ? dcnb('su-bg-saa-black', styles.membershipCardWrapper(disabled))
-            : styles.membershipCardWrapper(disabled),
-          newContactWrapper
+          styles.membershipCardWrapper,
+          newContact && 'su-border-dashed'
         )}
         onClick={toggleRelationship}
       >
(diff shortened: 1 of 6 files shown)


jdwjdwjdw (changes requested):
Overall looking great! A couple of things I noticed:

- The select/unselect functionality works as expected almost everywhere, but I did notice that if you click `Myself`, then `Pay in installments`, then go back and select `Someone else` (while keeping `Pay in installments` selected), the `Select membership type` button routs to https://deploy-preview-568--stanford-alumni.netlify.app/membership/register/installments/form instead of https://deploy-preview-568--stanford-alumni.netlify.app/membership/register/related-contacts. Would it be possible to have `Pay in installments` automatically be unselected if you switch to the Someone else card option?
- I'm seeing some issues/differences with the card styling on select:
  - For all of the cards, `su-bg-saa-black` should only be applied while selected
  - For the newContact card, `su-border-dashed` should only be applied while the card is not selected

src/components/page-types/membershipFormPage/membershipCard.js
```diff
@@ -41,23 +40,14 @@ const MembershipCard = ({
...
       <button
         type="button"
         className={dcnb(
-          isSelected
-            ? dcnb('su-bg-saa-black', styles.membershipCardWrapper(disabled))
-            : styles.membershipCardWrapper(disabled),
-          newContactWrapper
+          styles.membershipCardWrapper,
+          newContact && 'su-border-dashed',
+          isSelected && newContact ? 'su-bg-saa-black' : ''
```
rebeccahongsf, line 50: I encountered an issue with `dcnb` not applying the border styles within the `membershipCard.styles.js` file 🤔 (I'm guessing you encountered this as well in your membership card styles iteration.) I've reworked it into a conditional ternary operator 👍

rebeccahongsf:
@jdwjdwjdw ready for another round of feedback 🖖

src/components/page-types/membershipFormPage/membershipCard.js
jdwjdwjdw: yeah I definitely encountered some strange issues with that, but just attributed it to user error at the time 🤔

Thanks for the rework! Can you adjust the `newContact` card so the dashed border is removed/border solid added on select? I'm also only seeing `su-bg-saa-black` being applied to the `newContact` selected card, so that will need to be updated for the default membership cards as well.

jdwjdwjdw (changes requested):
Updated select functionality works great 👏 just some minor styling changes, then we should be good to go.

rebeccahongsf:
And we're ready for another round of feedback @jdwjdwjdw 🤞😄

PR #652, opened 2023-04-25
NOJIRA | @rebeccahongsf | add additional check for undefined data

rebeccahongsf:
# READY FOR REVIEW

# Summary
- add additional check for undefined data

# Review By (Date)
- asap; I'd like to push this fix to PROD as well since it is blocking people from signing up for T/S things

# Review Tasks

1. Navigate to https://deploy-preview-652--stanford-alumni.netlify.app/
2. Log in as _... rebecca is looking for a test account with no contact data_
3. Verify that you are able to log in without any issues and view your data at https://deploy-preview-652--stanford-alumni.netlify.app/api/auth/profile
4. Review code 🪴

### DIFF
--- a/src/api/auth/profile.js
+++ b/src/api/auth/profile.js
@@ -29,7 +29,10 @@ const megaprofileHandler = async (req, res, next) => {
   const resolved = await Promise.allSettled(requests);
 
   // Full GG Data.
-  if (resolved[0].status === 'fulfilled') {
+  if (
+    resolved[0].status === 'fulfilled' &&
+    !(resolved[0].value?.data === undefined)
+  ) {
     fullgg = resolved[0].value.data;
   } else {
     fullgg.name = {};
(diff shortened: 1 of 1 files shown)


src/api/auth/profile.js
```diff
@@ -29,20 +29,29 @@ const megaprofileHandler = async (req, res, next) => {
   const resolved = await Promise.allSettled(requests);
 
   // Full GG Data.
-  if (resolved[0].status === 'fulfilled') {
+  if (
+    resolved[0].status === 'fulfilled' &&
+    !(resolved[0].value?.data === undefined)
```
sherakama, line 34: @rebeccahongsf I think checking for any non-truthy response instead of directly checking for undefined would be better here. We will still have issues if the value of data is `null` or ``.

sherakama (approved):
Please adjust the check from checking for `undefined` to any non-truthy value.

sherakama (changes requested):
See prior comment

sherakama (approved):

sherakama:
Thanks @rebeccahongsf GTG.

PR #667, opened 2023-05-16
nojira | @rebeccahongsf | add date picker specific styles

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Add date picker specific styles

# Review By (Date)
- Whenever

# Review Tasks

1. Navigate to `/travel-study/destinations/italy-test/register/form`
2. Continue to the reg form
3. Fill out form and add a new contact
4. Verify that the DOB field is now styled like so:
image

5. Review code 🐛 

- I saw this while briefly looking into `SR01324343 `

### DIFF
--- a/src/styles/forms.css
+++ b/src/styles/forms.css
@@ -10,8 +10,7 @@
   @apply hocus:su-border-none su-text-black-30 su-bg-transparent su-p-0 su-type-1 su-font-semibold su-text-white su--mt-2;
 }
 
-.ggeWidget .ggeQuestion__field input[type="date"],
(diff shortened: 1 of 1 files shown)


rebeccahongsf:
> Updated date picker styles look good to me! Once we merge membership into dev, it looks like these fixes will update the date picker in membership as well, correct? Just want to make sure we address that in membership if needed, because I'm seeing the date picker when adding a new contact there as well.

Thanks for the speedy review, JDW! That's a good thing to highlight. This date field style update is global so the membership date picker should be styled as well. I'll make a note to double check once this PR is merged into dev (and we update the membership branch) 😃

PR #695, opened 2023-08-28
ADAPT-5902 | @rebeccahongsf | theme mega menu

rebeccahongsf:
# READY FOR REVIEW

# Summary
- theme mega menu
https://www.figma.com/file/nPKfSbRk5lXIzpDYGpS93G/Mega-Menu?type=design&node-id=12%3A476&mode=design&t=Ll9CTbMFY7op74rY-1

# Review By (Date)
- Retroreview :) 

# Review Tasks

1. Navigate to 
2. Verify that the mega menu appears as outlined in figma
3. Review code 

- ADAPT-5902
- DS-102


src/components/navigation/MegaMenu/megaMenuLinkItem.js
```diff
@@ -1,11 +1,18 @@
...
+import { dcnb } from 'cnbuilder';
 import SbLink from '../../../utilities/sbLink';
+import * as styles from './megaMenuLinkGroup.styles';
 
 const MegaMenuLinkItem = ({ blok: { link, linkText }, blok }) => (
   <SbEditable content={blok}>
-    <li>
-      <SbLink link={link}>{linkText}</SbLink>
+    <li className="su-rs-mb-0">
+      <SbLink
```
rebeccahongsf, line 10: Question for Tuesday @yvonnetangsu 😃 — Does Storyblok link field auto-magically define the external links differently from the internal links? I know we have a conditional to check within the  `SbLink` component but I was wondering if any additional code work is needed to make it work 🤔

yvonnetangsu: @rebeccahongsf I don't believe you need to do anything to the link prop, but I'm looking at the current navItem component, and there is a hasExternalLinkIcon prop that you can set (whether to display an external link icon) 😄 
https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/navigation/navItem.js

yvonnetangsu: It's more for whether you want all external links to display an icon - I don't think you need to do the external link check yourself.

yvonnetangsu:
@rebeccahongsf One thing I noticed is that you might be able to reuse more of the responsive styles for the top level parent buttons for smaller breakpoints. You're probably still working on it though? 😄 

PR preview:
Screenshot 2023-08-29 at 12 27 19 PM

On live site:
Screenshot 2023-08-29 at 12 19 00 PM

rebeccahongsf:
> @rebeccahongsf One thing I noticed is that you might be able to reuse more of the responsive styles for the top level parent buttons for smaller breakpoints. You're probably still working on it though? 😄
> 
> PR preview: Screenshot 2023-08-29 at 12 27 19 PM
> 
> On live site: Screenshot 2023-08-29 at 12 19 00 PM

Ooo! This is a great call out, @yvonnetangsu! 😅 I totally overlooked the parent text second line field

yvonnetangsu:
> Ooo! This is a great call out, @yvonnetangsu! 😅 I totally overlooked the parent text second line field

I figured you were probably still working on it too - there is no rush! 😄

yvonnetangsu:
@rebeccahongsf Very nice work - this is 95% there 😄 ! Here comes just a few comments then should be GTG

- [ ] On LG and +, when a mega menu panel is open, please add functionality to close with the escape key
- [ ] Please ask designer for a hocus state for this close button 😄 

Screenshot 2023-08-30 at 10 05 40 AM

- [ ] For Travel Study, there seems to be a visual regression which cause the normal state of the close button to become black, and the hocus state is taking on the one for homesite as well

Screenshot 2023-08-30 at 10 05 58 AM

Screenshot 2023-08-30 at 10 08 01 AM

- [ ] This is a designer thing - I would push back hard on this layout especially for the XL breakpoint. It really seems like esp for the panels with the card, it would take on the full width of the container here

Screenshot 2023-08-30 at 9 55 22 AM

- [ ] Also, another thing to check with designer - since the masthead of the SAA homesite doesn't quite follow the normal center container (ie, the screen margins are a bit different to accommodate the logo treatment, I'd ask the designer whether to align the content container to this modified container, eg, on the left edge, have the first column in the panel left aligned with the left side of the logo.

- [ ] I would suggest finetuning the width/flex treatment of the individual column a bit (maybe best to use a CSS grid here), eg, currently if I have a long item in a column it would cause the other column to shrink

Screenshot 2023-08-30 at 10 14 29 AM

There are different ways to update this, using a CSS grid might be the most straightforward, but you can also add su-shrink-0 to each column and give them each a max width etc. Up to you and the designer 😄

yvonnetangsu (changes requested):
Just added some comments for final finetuning - looking really great! Nice job 👏🏼 😄

rebeccahongsf:
@yvonnetangsu I updated this PR with some of your requested changes.

I'm going to follow up with a designer and implement the following changes in my next PR. Please let me know if any additional adjustments are needed. Thank you thank you! 😃 

> * [ ]  Please ask designer for a hocus state for this close button 😄
> 
> Screenshot 2023-08-30 at 10 05 40 AM
> 
> * [ ]  For Travel Study, there seems to be a visual regression which cause the normal state of the close button to become black, and the hocus state is taking on the one for homesite as well
> 
> Screenshot 2023-08-30 at 10 05 58 AM
> 
> Screenshot 2023-08-30 at 10 08 01 AM
> 
> * [ ]  Also, another thing to check with designer - since the masthead of the SAA homesite doesn't quite follow the normal center container (ie, the screen margins are a bit different to accommodate the logo treatment, I'd ask the designer whether to align the content container to this modified container, eg, on the left edge, have the first column in the panel left aligned with the left side of the logo.

yvonnetangsu:
Looks great - @rebeccahongsf !!

Adding that comment for grid gap as you requested - here's a screenshot - need gaps on both of the grids 👍🏼 😄 
Screenshot 2023-09-01 at 1 19 17 PM

yvonnetangsu:
A few more that I just noticed - related to the modal visual regression 😂 Definitely related to the earlier comments, so you might fix these when you fix those ones.

- [ ] Looks like the mobile user menu has some visual regression with the links/buttons
Screenshot 2023-09-01 at 1 22 24 PM

- [ ] Looks like the search modal Close X button is black so blending in (mobile and desktop)
Screenshot 2023-09-01 at 1 22 35 PM

yvonnetangsu:
I'm not able to logged in as zguan on the preview, if you could - please do a check for the user menu for me too 😄

rebeccahongsf:
> I'm not able to log in as zguan on the preview, if you could - please do a check for the user menu for me too 😄

Looks like I was missing the MEGAPROFILE_URL in vault! Mind retrying here?
https://megamenu--stanford-alumni.netlify.app/ 

I'll definitely take a look over the user menu as well 😃

yvonnetangsu:
> > I'm not able to log in as zguan on the preview, if you could - please do a check for the user menu for me too 😄
> 
> Looks like I was missing the MEGAPROFILE_URL in vault! Mind retrying here? https://megamenu--stanford-alumni.netlify.app/
> 
> I'll definitely take a look over the user menu as well 😃

Thank you @rebeccahongsf ! Yes, I can log in now 👏🏼 

Yup, looks like some visual regression in the user menu on the homesite after logging in as well
Screenshot 2023-09-05 at 12 24 12 PM

On the T/S side, it looks good except for the close modal button 😄 (for user menu, search - as mentioned before, and for the Destination page filter modal on mobile)
Screenshot 2023-09-05 at 12 25 15 PM

Screenshot 2023-09-05 at 12 27 54 PM

rebeccahongsf:
Thank you thank you @yvonnetangsu for looking it over again! Looks like my changes to the Modal.styles.ts is causing those issues so I'll fix that up in my mobile pr 🫡

yvonnetangsu:
Thank you @rebeccahongsf !! 😄 👍🏼

PR #696, opened 2023-09-05
DS-173 | @rebeccahongsf | theme mobile mega menu

rebeccahongsf:
# READY FOR REVIEW

# Summary
- theme mobile mega menu to use the same colors as the Giving site

# Review By (Date)
- When possible

# Review Tasks

1. Navigate to https://deploy-preview-696--stanford-alumni.netlify.app/
2. Verify that the mobile menu uses the same colors and spacing as Giving
3. Verify that the desktop menu is not impacted by these changes
4. Review code 🌱 

- [DS-173](https://stanford.atlassian.net/browse/DS-173)

### DIFF
--- a/src/components/navigation/MegaMenu/megaMenuLinkGroup.js
+++ b/src/components/navigation/MegaMenu/megaMenuLinkGroup.js
@@ -12,7 +12,7 @@ const MegaMenuLinkGroup = ({
   onlyLinks,
 }) => (
   <SbEditable content={blok}>
-    <GridCell lg={3}>
+    <GridCell lg={3} className={styles.menuGroup}>
       {heading && (
         <SbLink link={secondaryLink} classes={styles.menuLink}>
           <Heading weight="bold" uppercase className="su-text-17 su-rs-mb-0">
--- a/src/components/navigation/MegaMenu/megaMenuLinkGroup.js
+++ b/src/components/navigation/MegaMenu/megaMenuLinkGroup.js
@@ -2,23 +2,16 @@ import React from 'react';
 import SbEditable from 'storyblok-react';
 import CreateBloks from '../../../utilities/createBloks';
 import { Heading } from '../../simple/Heading';
-import SbLink from '../../../utilities/sbLink';
 import * as styles from './megaMenuLinkGroup.styles';
 import { GridCell } from '../../layout/GridCell';
 
-const MegaMenuLinkGroup = ({
-  blok: { heading, secondaryLink, links },
-  blok,
-  onlyLinks,
-}) => (
+const MegaMenuLinkGroup = ({ blok: { heading, links }, blok, onlyLinks }) => (
   <SbEditable content={blok}>
     <GridCell lg={4} className={styles.menuGroup}>
       {heading && (
-        <SbLink link={secondaryLink} classes={styles.menuLink}>
-          <Heading weight="bold" uppercase className="su-text-17 su-rs-mb-0">
-            {heading}
-          </Heading>
-        </SbLink>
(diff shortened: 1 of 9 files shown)


yvonnetangsu (changes requested):
Thank you for the changes, @rebeccahongsf ! We're 98% there 😄 Just a couple things

- [x] This one is a really small one. On mobile only, if you could please make the "drop down" relative positioned, so that when a dropdown is expanded, it doesn't cover the item below

Before:
Screenshot 2023-09-06 at 9 52 09 AM

After:
Screenshot 2023-09-06 at 9 52 25 AM

- [ ]  This one might need some designer input. On Giving, the card occupies a larger width than the text link columns. On the LG breakpoint, this makes a difference. Currently, in the mega menu preview, it looks like the card expands when there's a longer word in the card and causes overflow. It seems that it could use a larger width on LG.

For example, for the same browser width, the 1st card is skinnier because there are longer words in the 2nd card causing it to overflow 😄 
Screenshot 2023-09-06 at 9 39 04 AM
Screenshot 2023-09-06 at 9 39 25 AM

I see a few ways we can deal with this - one is for the card to take up a larger width. We could also reduce the padding, and font sizes for the card on LG (and add `su-break-words`). Perhaps a quick check with Bhavika 😄 I would test with the long word "Opportunities".

Let me know if you have any questions - almost there!! Nice job 👏🏼

rebeccahongsf:
Sounds good and thanks Yvonne! 😃 I went ahead to make the change for the mobile menu positioning and the word break for lg. I'll follow up with Bhavika tomorrow about the card width and word break to see what she recommend.

yvonnetangsu:
Thank you @rebeccahongsf I'll check back tomorrow 👍🏼 😄

src/components/navigation/MegaMenu/megaMenuPanel.js
```diff
@@ -99,21 +99,18 @@ const MegaMenuPanel = ({
...
+                  <CreateBloks blokSection={linkGroups} />
                 </Grid>
                 <div className="su-rs-mt-4">
                   <CreateBloks blokSection={sectionCtaLink} />
                 </div>
               </GridCell>
-              {card.length > 0 && (
-                <GridCell lg={3}>
-                  <CreateBloks blokSection={card} />
+              {fourthCol.length > 0 && (
```
rebeccahongsf, line 111: @yvonnetangsu I've confirmed with Bhavika about making the card span the same widths as Online Giving. 

As part of this update, I changed the card specific columns to be a general fourth columns where the user can either add a card or group of links. 

I figured this would be easiest to just have the start of the fourth column hardcoded (versus trying to calculate the sub-grid's width when there's 4 column of links versus a 3 column of links and card combo).

I'll most likely roll this in today and start some accessibility review but please continue to let me know your thoughts and if any additional adjustments should be made. I can always open up another PR for any other tweaks! Thank you 😄

src/components/navigation/MegaMenu/megaMenuLinkGroup.js
```diff
@@ -2,23 +2,16 @@ import React from 'react';
...
 
-const MegaMenuLinkGroup = ({
-  blok: { heading, secondaryLink, links },
-  blok,
-  onlyLinks,
-}) => (
+const MegaMenuLinkGroup = ({ blok: { heading, links }, blok, onlyLinks }) => (
   <SbEditable content={blok}>
-    <GridCell lg={3}>
+    <GridCell lg={4} className={styles.menuGroup}>
```
rebeccahongsf, line 10: I removed the link for the group header since it doesn't look like they will be using that functionalty (based on their link(s) spreadsheet) 😃

PR #708, opened 2023-09-11
DS-40 | @rebeccahongsf | Alumni Homesite Mega Menu

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Build out Alumni Homesite Mega Menu

# Review By (Date)
- When possible

# Review Tasks

Time can be tracked to [DS-202](https://stanford.atlassian.net/browse/DS-202)
1. Navigate to https://megamenu--stanford-alumni.netlify.app/
2. Verify that the mega menu matches the [Figma mocks](https://www.figma.com/file/nPKfSbRk5lXIzpDYGpS93G/Mega-Menu?type=design&node-id=12%3A476&mode=design&t=MlK6Ut7WqTPwubU7-1)
3. Review microsite(s) nav and confirm that it has not been impacted by this change
https://megamenu--stanford-alumni.netlify.app/travel-study/
4. Review code

- [DS-40](https://stanford.atlassian.net/browse/DS-40)
- [DS-202](https://stanford.atlassian.net/browse/DS-202)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# [ADAPT SAA Alumni: Mega Menu](https://github.com/SU-SWS/saa_alumni)
+# [ADAPT SAA Alumni](https://github.com/SU-SWS/saa_alumni)
  
 <a href="https://codeclimate.com/github/SU-SWS/saa_alumni/maintainability"><img src="https://api.codeclimate.com/v1/badges/6545d64d90025f75e2c8/maintainability" /></a>
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# [ADAPT SAA Alumni](https://github.com/SU-SWS/saa_alumni)
+# [ADAPT SAA Alumni: Mega Menu](https://github.com/SU-SWS/saa_alumni)
  
 <a href="https://codeclimate.com/github/SU-SWS/saa_alumni/maintainability"><img src="https://api.codeclimate.com/v1/badges/6545d64d90025f75e2c8/maintainability" /></a>
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -47,7 +47,7 @@
     endpoint = 'https://vault.stanford.edu'
     # Order doesn't matter here as these will be fetched asyncronously.
     # Try not to have dupes.
-    secrets = ['secret/data/projects/adapt/saml', 'secret/data/projects/adapt/megamenu']
+    secrets = ['secret/data/projects/adapt/saml', 'secret/data/projects/adapt/alumni']
     # Set which environment variables to use for the approle login
     secretId = 'VAULT_SECRET_ID'
     roleId = 'VAULT_ROLE_ID'
--- a/netlify.toml
+++ b/netlify.toml
@@ -47,7 +47,7 @@
     endpoint = 'https://vault.stanford.edu'
     # Order doesn't matter here as these will be fetched asyncronously.
     # Try not to have dupes.
-    secrets = ['secret/data/projects/adapt/saml', 'secret/data/projects/adapt/alumni']
+    secrets = ['secret/data/projects/adapt/saml', 'secret/data/projects/adapt/megamenu']
     # Set which environment variables to use for the approle login
     secretId = 'VAULT_SECRET_ID'
     roleId = 'VAULT_ROLE_ID'
--- a/src/components/components.js
+++ b/src/components/components.js
@@ -84,6 +84,11 @@ import VerticalNavWrapper from './navigation/verticalNavWrapper';
 import VerticalNavItem from './navigation/verticalNavItem';
 import Wysiwyg from './simple/wysiwyg';
 import MembershipPaymentOptions from './page-types/membershipFormPage/membershipPaymentOptions';
+import MegaMenu from './navigation/MegaMenu/megaMenu';
+import MegaMenuPanel from './navigation/MegaMenu/megaMenuPanel';
+import MegaMenuCard from './navigation/MegaMenu/megaMenuCard';
+import MegaMenuLinkItem from './navigation/MegaMenu/megaMenuLinkItem';
+import MegaMenuLinkGroup from './navigation/MegaMenu/megaMenuLinkGroup';
 
 const ComponentList = {
   accordion: Accordion,
@@ -126,6 +131,11 @@ const ComponentList = {
   mainNavItem: MainNavItem,
   masthead: Masthead,
   mastheadPicker: MastheadPicker,
+  megaMenu: MegaMenu,
+  megaMenuPanel: MegaMenuPanel,
+  megaMenuCard: MegaMenuCard,
+  megaMenuLinkItem: MegaMenuLinkItem,
+  megaMenuLinkGroup: MegaMenuLinkGroup,
   membershipCard: MembershipCard,
   membershipFormPage: MembershipFormPage,
   membershipFullPaymentForm: MembershipFullPaymentForm,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "saa-alumni",
-  "version": "4.0.8",
+  "version": "4.0.9",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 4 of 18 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# [ADAPT SAA Alumni](https://github.com/SU-SWS/saa_alumni)
+# [ADAPT SAA Alumni: Mega Menu](https://github.com/SU-SWS/saa_alumni)
```
rebeccahongsf, line 1: @rebeccahongsf revert before merge

netlify.toml
```diff
@@ -47,7 +47,7 @@
     endpoint = 'https://vault.stanford.edu'
     # Order doesn't matter here as these will be fetched asyncronously.
     # Try not to have dupes.
-    secrets = ['secret/data/projects/adapt/saml', 'secret/data/projects/adapt/alumni']
+    secrets = ['secret/data/projects/adapt/saml', 'secret/data/projects/adapt/megamenu']
```
rebeccahongsf, line 50: @rebeccahongsf revert before merge

sherakama:
Tagging in @ericbakenhus for review and exposure.

yvonnetangsu:
I like the new adjustment of the card width etc @rebeccahongsf ! Could we replace the su-w-fit class with the su-w-full class so the card always take up the full width of the slot that it's in, eg, before and after.

Screenshot 2023-09-12 at 9 56 26 AM
Screenshot 2023-09-12 at 9 57 22 AM

yvonnetangsu:
I'll take another look after a couple meetings this morning 😄 Thank you for the updates!

yvonnetangsu (changes requested):
@rebeccahongsf Let me know if I'm looking at the latest. I logged in and seeing this in the mobile user menu 😄 The logged out state looks ok though 👍🏼 

Screenshot 2023-09-12 at 12 59 38 PM

yvonnetangsu:
@rebeccahongsf Did another round of visual QA and all looks good to me! 🎉 😄 I really like the new decision to go with the wider width of the 4th column. I have 3 very minor things - please feel free to do later and no need to block this from merging since I know you're also trying to get a SODA review. I'll post them below:

- Could we change the `display` of the links to `inline-block`, and use a line-height of 1.2 to tighter up the space a bit when they're multilined? Before and after:
Screenshot 2023-09-12 at 12 51 08 PM
Screenshot 2023-09-12 at 12 50 58 PM

- I'm seeing a 1px gap for the cards sometimes (it's a browser glitch thing that I've dealt with previously). One quick fix is to just shift the red part 1px up.
Screenshot 2023-09-12 at 12 57 34 PM

- Is it possible to change the background color of the main menu modal to all white? When it's black like how it currently is, I found myself trying to click on the black portion to click out of the menu (it gave me the illusion that the black portion is not part of the menu) - before and after:
Screenshot 2023-09-12 at 12 58 56 PM

Screenshot 2023-09-12 at 12 59 14 PM


I noticed that the search modal "x" button is missing a hocus state - but that was visual regression from before so not related to this PR. I'll take care of that in my Decanter update ticket 👍🏼 😄

rebeccahongsf:
Thanks @yvonnetangsu! 😃 I added the requested fixes and an accessibility fix (the rest were content authoring related). I think we might let the client review the mega menu preview site. It depends if they want access to review the SB authoring side of things. We'll see during sprint planning today 🤔

ericbakenhus (approved):
Looks really awesome! 🦖

ralph-wiggum-simpsons

rebeccahongsf:
> Looks really awesome! 🦖
> 

Thanks @ericbakenhus!! 😆 Your help is greatly appreciated!

yvonnetangsu:
> Looks really awesome! 🦖
> 
> ralph-wiggum-simpsons

Just seeing your presence helps @ericbakenhus 😆 👋🏼

src/components/components.js
```diff
@@ -126,6 +131,11 @@ const ComponentList = {
   mainNavItem: MainNavItem,
   masthead: Masthead,
   mastheadPicker: MastheadPicker,
+  megaMenu: MegaMenu,
+  megaMenuPanel: MegaMenuPanel,
+  megaMenuCard: MegaMenuCard,
+  megaMenuLinkItem: MegaMenuLinkItem,
+  megaMenuLinkGroup: MegaMenuLinkGroup,
```
sherakama, line 138: @rebeccahongsf Are any of these new storyblok page or component types? Do you need to remove them from the GraphQL query and/or search?

sherakama:
Ditto to the other comments. The rest of this looks great. Keyboard nav and focus looks good too.

src/components/components.js
rebeccahongsf: Ooo! Great questions @sherakama 🤔 These are indeed new nestable blok types that are used within the masthead component. Since they are not universal bloks, I don't think we need to exclude them from graphQL query, search, and/or XML sitemap. 

I believe we should be covered with omitting the `masthead` from the GraphQL query here: 
https://github.com/SU-SWS/saa_alumni/blob/dev/gatsby-node.js#L16

Any thoughts or concerns? 👀

sherakama: Ok, that sounds good. As long as they don't show up I am happy.

The last thing is that the build still builds if all of the items are unpublished/deleted

See: https://github.com/SU-SWS/saa_alumni/blob/dev/gatsby-node.js#L362

sherakama (approved):

src/components/components.js
rebeccahongsf: Ah yes, another good thing to triple check! 😉 

The build will still build even if the menu link/page is unpublished/deleted. The link will still appear in the menu and users accessing the link will be redirect to the 404 err page in the event that the menu link/page is unpublished/deleted.

sherakama: I wasn't so worried about the destination of the link but that is a good check.

I was more worried about the situation the `createSchemaCustomization` fixes. This is for the situation where we are pulling content using GraphQL but it is unable to infer the field types because all of the content is unpublished or deleted so there is nothing to infer from. This would be for the new blocks `MegaMenu***` and their fields. I understand it will be unlikely for these to be completely unpublished in the long term but what is the go-live strategy with this content? Are you planning to let them build out an unpublished version first? If so, you might want to check what happens when all of the MegaMenu*** blocks are unpublished and you're not using the Storyblok preview token.

rebeccahongsf: > I understand it will be unlikely for these to be completely unpublished in the long term but what is the go-live strategy with this content? Are you planning to let them build out an unpublished version first? 

Yes! As part of our enhancement, SAA has requested that we assemble the mega menu on prod. I will be creating a new masthead (with the `MegaMenu***` blocks) for them to review. Once confirmed, I'll duplicated the swap out the live `MainMenu***` block with the approved `MegaMenu***` block. 

That way we do not need to go into each page to update the Masthead.

> If so, you might want to check what happens when all of the `MegaMenu***` blocks are unpublished and you're not using the Storyblok preview token.

Is the concern around an unpublished Masthead containing the `MegaMenu***` blocks? 

When a Masthead is unpublished, it doesn't display the entire masthead block on the page.
Example: https://megamenu--stanford-alumni.netlify.app/test/rebecca/rehong-masthead/

Is there a way to unpublish specific nested blocks? 🤔

sherakama: I just tested and I think you're all good here.

What I did is 
1. pull your latest code from this branch
2. Change the STORYBLOK_ACCESS_TOKEN to point to the production space
3. Run `npm run build`

The build was successful so you are GTG.

yvonnetangsu: Is this ready to merge into `dev` @rebeccahongsf and @sherakama ? I didn't touch the menu/modal components in my Decanter update PR to avoid ugly merge conflicts, so after this get merged, I can get this into my PR and work on cleaning those up next week 😄

rebeccahongsf: @yvonnetangsu oh yes, this PR is essentially ready! Business is QAing the megamenu preview site but I can always revert the files back after merge and open a new PR if they have any additional requests 😃

I'll get this PR fixed up and merged in this morning

PR #750, opened 2023-11-08
DS-265 | @rebeccahongsf | fixup nonce error

rebeccahongsf:
# READY FOR REVIEW

# Summary
- fixup nonce error

# Review By (Date)
- Retroreview. Rolling this in so I can have a working demo for tomorrow's meeting with membership. I'll be implementing the error use cases in a separate PR and will roll up any feedback into that work. Thank you! :D

# Review Tasks

1. Checkout branch on your local
2. Enable mockserver
3. Navigate to /membership/payment-history
4. Verify that the self-service portal loads
5. Navigate to membership/payment-update
6. Verify that the self-service portal loads
7. Review code

- DS-265

### DIFF
--- a/src/components/page-types/lightFormPage/lightFormPage.js
+++ b/src/components/page-types/lightFormPage/lightFormPage.js
@@ -35,7 +35,21 @@ const LightFormPage = (props) => {
   const [kwoCreds, setKwoCreds] = useState('');
   const memberships = userProfile?.memberships;
   const [loading, setLoading] = useState(true);
-  const [error, setError] = useState(true);
+  const [error, setError] = useState('');
+
+  let paymentRefId = false;
+
+  if (memberships && loading) {
+    memberships.forEach((membership) => {
+      if (
+        !paymentRefId &&
+        membership.membershipGroup === 'SAA' &&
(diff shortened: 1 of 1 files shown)


src/components/page-types/lightFormPage/lightFormPage.js
```diff
@@ -35,56 +35,58 @@ const LightFormPage = (props) => {
   const [kwoCreds, setKwoCreds] = useState('');
   const memberships = userProfile?.memberships;
   const [loading, setLoading] = useState(true);
-  const [error, setError] = useState(true);
+  const [error, setError] = useState('');
+
+  let paymentRefId = false;
```
rebeccahongsf, line 40: This works but just because it works doesn't mean it's right 🤔 There are a few use cases that I need to account for:
- Not having membership
- Having a membership but no payment ref id yet (in this case I assumed `membershipGGPaymentReferenceID: null` since that is how it appears)
- Having a membership but no installment plan since they paid in full

I'll revisit this logic in [DS-433 | Configure no payment ref id error messages](https://stanford.atlassian.net/browse/DS-433). Super open to any thoughts and feedback in the meantime!

PR #752, opened 2023-11-10
DS-160 | @rebeccahongsf | build gsb membership card

rebeccahongsf:
# READY FOR REVIEW

# Summary
- build gsb membership card

# Review By (Date)
- When possible

# Review Tasks

1. Navigate to https://deploy-preview-752--stanford-alumni.netlify.app/membership/gsbcard
2. Verify that the GSB card appears with the degree cert information and appears similarly to the [SAA Card](https://deploy-preview-752--stanford-alumni.netlify.app/membership/gsbcard)
image

3. Review code

- [DS-160](https://stanford.atlassian.net/browse/DS-160)


ericbakenhus:
@rebeccahongsf Comparing the SAA card page the GSB card page I do see a few minor differences:

- It looks like the card/content area has a different min width on both pages, which makes the GSB card content seem further to the left on wider screens
- The SAA page has an additional "Your card" H2
- The SAA card has a gradient background and the GSB card doesn't

I can imagine some/all of those were intentional decisions though, so feel free to ignore if they were. 😅

src/components/cards/gsbCard.js
```diff
@@ -1,92 +1,86 @@
...
+                <div className="su-relative su-overflow-hidden su-rounded-[3rem] sm:su-w-[520px] su-mb-50 sm:su-mb-90 lg:su-mb-0 su-text-white su-bg-cardinal-red-light">
+                  <div className="su-relative su-w-full su-pt-[63%]">
+                    <div className="su-absolute su-top-0 su-w-full su-h-full">
+                      <div className="su-relative su-flex su-flex-col su-h-full su-flex su-text-[38px] su-z-10 su-pl-[5%]">
+                        <div className="su-top-0 su-left-0 su-flex su-items-center su-w-[63%] su-h-[35%]">
+                          <img
+                            src={GsbLogoWhite}
+                            alt=""
+                            className="su-max-w-full su-max-h-full"
+                          />
```
ericbakenhus, line 41: I can understand the reasoning for considering this (and the existing SAA card logo) a presentational logo image, but it's also just an image of text. That makes me think we should probably go with the [Images of Text](https://www.w3.org/WAI/tutorials/images/textual/) guidelines and put all the words into the alt text.

If it was already run by SODA then obviously go with their recommendation instead. 😄

rebeccahongsf:
> @rebeccahongsf Comparing the SAA card page the GSB card page I do see a few minor differences:
> 
> * It looks like the card/content area has a different min width on both pages, which makes the GSB card content seem further to the left on wider screens
> * The SAA page has an additional "Your card" H2
> * The SAA card has a gradient background and the GSB card doesn't
> 
> I can imagine some/all of those were intentional decisions though, so feel free to ignore if they were. 😅

Thank you for taking a look and the speedy review, @ericbakenhus! These are all great points! I realize I didn't provide enough guidance and context in my testing instructions. There were intentional design decisions for sure.

With some recent feedback from the client, I did a rework of the GSB Card from a nestable block to a content type on Storyblok since they would like to include the Stanford GSB logo as well (and remove the masthead/footer). Mind taking another look? 😄

src/components/cards/gsbCard.js
stfurkan: @rebeccahongsf I'm not too familiar with Gatsby but should we use `gatsby-plugin-image` for image optimization?

https://www.gatsbyjs.com/plugins/gatsby-plugin-image/

rebeccahongsf: That's a fantastic point, @stfurkan! If we're considering wider use of  across multiple areas, it might be worth planning ahead and using that plugin. However, for these occasional instances of , leaving it as-is for now could be reasonable. What do you think about this approach? 😃

sherakama: The other option would be to upload the image to Storyblok's assets and then request it through their image optimization / cloudinary service using the https://github.com/SU-SWS/saa_alumni/blob/dev/src/utilities/transformImage.js utility.

sherakama: But it is _fine_ to have it as a raw img tag.

rebeccahongsf: Sounds good @stfurkan! I've created [DS-466](https://stanford.atlassian.net/browse/DS-466) for our backlog so we can update all the static `` to use the `gatsby-plugin-image`.

stfurkan (approved):
LGTM! Thanks @rebeccahongsf :)

PR #37, opened 2021-06-10
ADAPT-2066: Global and Local Alerts

romakravchuk:
# NOT READY FOR REVIEW
- (Edit the above to reflect status)

# Summary
- TL;DR - what's this PR for?

# Review By (Date)
- When does this need to be reviewed by?

# Criticality
- How critical is this PR on a 1-10 scale? Also see [Severity Assessment](https://stanfordits.atlassian.net/browse/D8CORE-1705).
- E.g., it affects one site, or every site and product?

# Review Tasks

## Setup tasks and/or behavior to test

1. Check out this branch
2. Navigate to...
3. Verify...

## Front End Validation
- [ ] Is the markup using the appropriate semantic tags and passes HTML validation?
- [ ] Cross-browser testing has been performed?
- [ ] Automated accessibility scans performed?
- [ ] Manual accessibility tests performed?
- [ ] Design is approved by @ user?

### Code
- [ ] Are the naming conventions following our standards?
- [ ] Does the code have sufficient inline comments?
- [ ] Is there anything in this code that would be hidden or hard to discover through the UI?
- [ ] Are there any [code smells](https://blog.codinghorror.com/code-smells/)?
- [ ] Are tests provided? eg (unit, behat, or codeception)

### Code security
- [ ] Are all [forms properly sanitized](https://www.drupal.org/docs/8/security/drupal-8-sanitizing-output)?
- [ ] Any obvious [security flaws or new areas for attack](https://www.drupal.org/docs/8/security)?

## General
- [ ] Is there anything included in this PR that is not related to the problem it is trying to solve?
- [ ] Is the approach to the problem appropriate?

# Affected Projects or Products
- Does this PR impact any particular projects, products, or modules?

# Associated Issues and/or People
- JIRA ticket(s)
- Other PRs
- Any other contextual information that might be helpful (e.g., description of a bug that this PR fixes, new functionality that it adds, etc.)
- Anyone who should be notified? (`@mention` them here)

# Resources
- [AMP Tool](https://stanford.levelaccess.net/index.php)
- [Accessibility Manual Test Script](https://docs.google.com/document/d/1ZXJ9RIUNXsS674ow9j3qJ2g1OAkCjmqMXl0Gs8XHEPQ/edit?usp=sharing)
- [HTML Validator](https://validator.w3.org/)
- [Browserstack](https://live.browserstack.com/dashboard) and link to [Browserstack Credentials](https://asconfluence.stanford.edu/confluence/display/SWS/External+Account+Credentials)

PR #757, opened 2023-11-16
NoJira | @shc314 | Adds two Hero components in red and blue color gradients

shc314:
# READY FOR REVIEW

# Summary
- Hero component has been duplicated so that one has a Brick gradient and the other has a gradient in Palo Alto dark.

# Review By (Date)
- When possible

# Review Tasks

1. Check out this branch
2. Place the heroBrick and heroPaloAltoDark blocks on the homepage
3. Verify that the colors appear according to the mockup provided by the designer

## Front End Validation

### Code

### Code security

## General

- HOM-1


rebeccahongsf (changes requested):
Hello @sch314! 😃  Great job on the gradient hero components. The gradient heros look terrific! I do have some suggestions to enhance the components in line with the DRY (Don't Repeat Yourself) principle.

Given the similarity between this component and the original hero components, I propose incorporating gradients as an option. Similar to how we offer Align and Button Size options for the CTA Button Link (found under Hero > CTA Button Link > Options tab), consider adding a 'Gradient' option in the existing Hero.

image

To do this, add a Single-Option field labeled 'Gradient' under the Image tab within the existing Hero component in Storyblok. 

image

This field would include options like:

- No Gradient
- Brick
- Palo Alto Dark

Each option should have its corresponding machine name (e.g., none, brick, palo-alto-dark).

As for handling this in code, the StoryBlok machine name can be added to the blok destructure list in the [`hero.js`](https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/composite/hero.js#L23) file. 

Within the same file, we'll also want to update the conditional logic for the gradients. [Line #37 - #39](https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/composite/hero.js#L37-L39) has an existing `isDarkGradient` conditional that can be expand on to include the new Single-Options machine name to determine which gradient to apply to the hero component.

Here's a rough example using a switch case (if/else conditions work as well if that is you preferred method!):

```
  let gradientFrom = 'su-to-saa-black';
  switch (heroGradient) { // Replace heroGradient with the single-option machine name
    case 'brick':
      gradientFrom = 'su-to-brick';
      break;

    case 'palo-alto-dark':
      gradientFrom = 'su-to-palo-alto-dark';
      break;

    default:
      gradientFrom = 'su-to-saa-black';
  }

  if (isDarkGradient) {
    gradientFrom = 'su-to-saa-black su-from-black-true-opacity-20';
  }
```

Where `su-from-transparent` replaces `su-to-saa-black` on [Line 62](https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/composite/hero.js#L62). This is one of the many approaches to implement this feature. Feel free to experiment with different methods for applying styles conditionally.

You might notice in my example, I left the if conditional for the `isDarkGradient` as-is. Adding the darker gradient as a selectable option in Storyblok, along with a corresponding switch case in the code, is definitely feasible. However, implementing this change would require a content author to manually update all existing Heros where the `isDarkGradient` boolean is active. I would defer to @William-Misiaszek for direction on that.

Give this a try, and if you have any questions, please don't hesitate to reach out. Thanks for your hard work, Shirley! 🥳

shc314:
Thank you for your feedback, @rebeccahongsf ! I have implemented your suggestions.

I wonder if the gradientFrom variable name can be confusing because now its values are actually gradient-to, but I'm not sure if changing it will affect anything else. It only appears in the hero.js file. Please let me know if you think it'll be better to change it to gradientTo or to leave it as is. Thank you!

rebeccahongsf:
Thank you so much for implementing my feedback, @shc314 ! The gradient hero updates are looking great 🎉 

As for your questions:

> 1. I wonder if the gradientFrom variable name can be confusing because now its values are actually gradient-to, but I'm not sure if changing it will affect anything else. It only appears in the hero.js file. Please let me know if you think it'll be better to change it to gradientTo or to leave it as is.

Great catch! I totally overlooked that. Please feel free to update the gradientFrom variable name to `gradientTo` or another descriptive variable that you see fit.

> 2. I just realized that I did not import heroGradient from dataSource. Would it be better to use the values defined there in the switch statement?

Another great question! My apologies for the confusion. If we were using the `Self` Source type in StoryBlok, the current switch statement works since the `Self` Source type is perfect for one time uses.
image

However, since we are using the `Datasource` Source type in StoryBlok, we can remove the switch statement, import the heroGradient options from the `dataSource.js`, and update the `gradientFrom` (or new gradient variable) as

```
import { heroGradient } from '../../dataSource.js' // double check the import since I guesstimated this :) 
...

const Hero = ({
  blok: {
    headlineSize,
    isDarkGradient,
    heroGradient: heroGradientProp, // redefine variable name so that it doesn't conflict with heroGradient from dataSource
    ...
  },
  blok,
}) => {
...

  let gradientFrom = heroGradient[heroGradientProp]

  if (isDarkGradient) {
    gradientFrom = 'su-from-black-true-opacity-20';
  }
```

I find [SAALinkButton.jsx](https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/cta/SAALinkButton.jsx#L50) helpful for referencing when working on new dataSources 😃 

> 3. The designer also wants to add a white gradient, but I could not find a predefined class for that. Could you tell me if there is one, and if not, where can I put it?

I believe the tailwind utility class for white would be `su-to-white`. Give that a try and if you encounter any issues, let me know! 🤔 

Feel free to tag me for further queries or additional information. Thanks again for your work on this!

shc314:
Thanks for your explanations, @rebeccahongsf ! I have a few more questions.

The designer would like the gradient colors reversed for the xs breakpoint. For the larger breakpoints, the gradient starts transparent at the top and the color is at the bottom. For the xs breakpoint, the color starts at the top with 30% opacity and then goes to #181D1C 100% for all three of them. Since the "to" color is the same for all three of them, I only put the "from" classes that change in the data source. I wonder if doing it this way will be confusing. Now `gradientFrom` contains both "to" and "from" values, so I'm not sure what to call it.

I was unable to find any classes for brick, palo-alto-dark, and white with 30% opacity, so I put them in tailwind/theme/colors.js. I tried writing them like this:

```
brick: {
  opacity: {
    30: 'rgba(101, 28, 50, 0.3)',
  },
},
'palo-alto-dark': {
  opacity: {
    30: 'rgba(1, 66, 64, 0.3)',
  }
},
white: {
  opacity: {
    30: 'rgba(255, 255, 255, 0.3)',
  }
},
```
but got compilation errors saying `su-bg-white` and `su-text-white` do not exist, so I defined the classes differently. It works but appears inconsistent with the way the other colors are defined. Please let me know if any of those classes already exist.

Thank you for all your help!

rebeccahongsf:
Hello and happy holidays, @shc314! Not a problem at all.

Since `gradientFrom` now contains both `to` and `from` values, we could consider renaming the variable to `gradientBackground` or simply `gradient`. I'll leave the final decision up to you! 😃 

There are various ways to incorporate opacities. Regarding the error you encountered with using white opacity when configuring colors, it might be due to the out-of-the-box Tailwind colors not being extended. A workaround could be adding the default white:

```
white: {
  DEFAULT: '#ffffff',
  opacity: {
    30: 'rgba(255, 255, 255, 0.3)',
  }
},
```
Then to use the opacity, it would be `su-from-white-opacity-30`.

Another method to define opacity without configuring new colors is by adding a forward slash followed by the opacity percentage to the end of the color. For instance, for 30% opacity white, you'd use `su-from-white/30`.

Hope this helps. Let me know if you have any other questions! 😊

shc314:
That was very helpful, @rebeccahongsf ! Thank you, and happy holidays!

You were correct about extending the white. After adding the default, the errors went away.
Since I'm not sure if those colors at 30% opacity will be reused, I decided to use the second option.

rebeccahongsf:
Huzzah! I'm thrilled to hear that it works, @shc314! 😊

Is your PR ready for another review? If so, please let us know and we'll gladly proceed with another code review.

I wanted to give you a heads up that I'll be starting my vacation tomorrow. When we return from the winter closure, I'll be transitioning off the Alumni Homesite project. To ensure a smooth transition, I'd like to introduce my wonderful colleagues, @stfurkan and @ericbakenhus. They'll be taking over to support you in future code reviews.

Furkan will be your primary contact, but feel free to tag both of them for feedback.

It has truly been a pleasure collaborating with you! Thank you again for your amazing work. Wishing you a cozy holiday season ☃️ 🎄

shc314:
Hi @rebeccahongsf and @stfurkan, my PR is ready for another code review.

Thank you for the update, Rebecca, and for all your help! Wishing you a wonderful vacation.

stfurkan:
@shc314 thanks for the great work :) When I add the `Hero` block to my test page, `Hero Gradient` option doesn't work for me (You can see the screenshots below). `Use darker gradient` toggle work and I can see the change on the page immediately.

Could you please fix it? If I'm testing it wrong, can you please let me know the correct way to test this change. Thanks again.

cc @sherakama 


image

image

sherakama:
> Could you please fix it? If I'm testing it wrong, can you please let me know the correct way to test this change. Thanks again.

The preview URL you are using in Storyblok is pointing to the dev site which doesn't have the code deployed that is being tested. Your choices are

1. Add a new URL that points to the preview build url
2. Pull the code down to localhost and run it, then select the localhost:3010 option from the preview list drop down

See: https://github.com/SU-SWS/saa_alumni#using-the-storyblok-editor-on-localhost

stfurkan:
> > Could you please fix it? If I'm testing it wrong, can you please let me know the correct way to test this change. Thanks again.
> 
> The preview URL you are using in Storyblok is pointing to the dev site which doesn't have the code deployed that is being tested. Your choices are
> 
> 1. Add a new URL that points to the preview build url
> 2. Pull the code down to localhost and run it, then select the localhost:3010 option from the preview list drop down
> 
> See: https://github.com/SU-SWS/saa_alumni#using-the-storyblok-editor-on-localhost

Thanks @sherakama :) I used deploy preview URL in Storyblok and it's working now. It seems like there are lots of things that I need to learn about Storyblok 


@shc314 LGTM! Thanks again for the work :) 

image

stfurkan (approved):

sherakama:
Thanks for the review @stfurkan 

@shc314, I'll update the branch with the latest from dev and when the tests all go green, I'll merge the work in to dev.

This will get into production, probably next year, when the next batch of work goes out. 

Thanks everyone.

sherakama:
@shc314,

I've also invited you to the production space so you can make the changes to the bloks there. I've copied over the datasource itself for you already.

sherakama:
Build completed in: https://app.netlify.com/sites/stanford-alumni/deploys/65809a72a6dfa70831b114cf

PR #804, opened 2024-02-06
DS-522: Update fontawesome v6

stfurkan:
# READY FOR REVIEW

# Summary
- Update fontawesome to  version 6

# Review By (Date)
- End of sprint

# Criticality
- 3

# Review Tasks

1. Check out this branch
2. Navigate to `alumni homesite` and make sure everything works as expected


package.json
```diff
@@ -6,12 +6,12 @@
...
-    "@fortawesome/free-brands-svg-icons": "^5.15.4",
-    "@fortawesome/free-solid-svg-icons": "^5.15.4",
-    "@fortawesome/pro-light-svg-icons": "^5.15.4",
-    "@fortawesome/pro-solid-svg-icons": "^5.15.4",
+    "@fortawesome/fontawesome-pro": "^6.5.1",
+    "@fortawesome/fontawesome-svg-core": "^6.5.1",
+    "@fortawesome/free-brands-svg-icons": "^6.5.1",
+    "@fortawesome/free-solid-svg-icons": "^6.5.1",
+    "@fortawesome/pro-light-svg-icons": "^6.5.1",
+    "@fortawesome/pro-solid-svg-icons": "^6.5.1",
```
stfurkan, line 14: fontawesome updates

package.json
```diff
@@ -22,7 +22,7 @@
     "fast-sort": "^3.4.0",
     "gatsby": "^4.25.7",
     "gatsby-link": "^4.25.0",
-    "gatsby-plugin-fontawesome-css": "^1.2.0",
+    "gatsby-plugin-fontawesome-css-2": "^1.3.0",
```
stfurkan, line 25: `gatsby-plugin-fontawesome-css` was causing npm dependency issues with fontawesome v6. I used the forked version of the package here.

See https://github.com/jzabala/gatsby-plugin-fontawesome-css/issues/22

sherakama: Is it just the one dependency conflict?

Cuz you could also go with [overriding the dependency](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) in package.json

Pros to that approach:
- You stay on the main repo and continue to get those updates

Cons to that approach:
- You still have to check in once and a while to see if you still need the override

stfurkan: Yes, currently it's only one dependency conflict. I don't think there is any value to stay on the main repo. [Last commit](https://github.com/jzabala/gatsby-plugin-fontawesome-css/commit/fc0898323033f1638da06a53cf7714340b9ee958) the main repo has was 3 years ago. I can try the `overriding the dependency` solution but I think in this case using the forked repo is good enough.

sherakama: Alright. I'm good with that.

sherakama (approved):
This is looking good to me.

@yvonnetangsu or @rebeccahongsf do you want to eyeball this quickly to see if you can pick up anything I missed?

yvonnetangsu (approved):
Looked at the T/S itinerary section which has the most number of icons and seems ok 😄 Did a quick check else where too. I think it's GTG 🎉  Thank you @stfurkan !

https://deploy-preview-804--stanford-alumni.netlify.app/travel-study/destinations/china-2024/#itinerary-section

stfurkan:
Thanks for the reviews @sherakama and @yvonnetangsu :)

PR #805, opened 2024-02-07
Task/hom 8 update referrer url code block

shc314:
# READY FOR REVIEW

# Summary
- Adds the referrepolicy attribute for the CTA button to be used with the Project Muse URL.

# Review By (Date)
- When possible

# Review Tasks

1. Check out this branch
2. Create a page with a CTA Button Link.
3. Under the Other Link Attributes tab, add "no-referrer-when-downgrade" in the referrerpolicy field. Confirm that it appears in the HTML for that link.
4.  Test other combinations with the rel attribute to make sure this change doesn't break the rel attribute functionality.

- JIRA ticket HOM-8
- To be used with the Project Muse button on this page https://alumni.stanford.edu/benefits/online-publication-databases/ (I couldn't find it on DEV)
- @William-Misiaszek

### DIFF
--- a/src/components/cta/SAALinkButton.jsx
+++ b/src/components/cta/SAALinkButton.jsx
@@ -21,6 +21,7 @@ const SAALinkButtonProps = {
   children: PropTypes.node,
   link: SBLinkType,
   rel: PropTypes.string,
+  referrerpolicy: PropTypes.string,
   attributes: PropTypes.oneOfType([PropTypes.object]),
   srText: PropTypes.string,
   className: ClassNameType,
@@ -36,6 +37,7 @@ export const SAALinkButton = React.forwardRef(
       children,
       link,
       rel,
+      referrerpolicy,
       attributes,
       srText,
       className,
@@ -57,7 +59,7 @@ export const SAALinkButton = React.forwardRef(
         <SbLink
           ref={ref}
           link={link}
-          attributes={attributes || rel ? { rel, ...attributes } : {}}
+          attributes={attributes || rel && referrerpolicy ? { rel, referrerpolicy, ...attributes } : rel ? {rel, ...attributes} : referrerpolicy ? {referrerpolicy, ...attributes} : {}}
           classes={dcnb(styles.link, ctaButtonStyle, ctaButtonSize, className)}
           {...rest}
         >
--- a/src/components/storyblok/ctaButton.jsx
+++ b/src/components/storyblok/ctaButton.jsx
@@ -3,7 +3,7 @@ import SbEditable from 'storyblok-react';
 import { SAALinkButton } from '../cta/SAALinkButton';
 
 export const SBCtaButton = ({ blok }) => {
-  const { size, buttonStyle, icon, align, linkText, link, rel, srText } = blok;
+  const { size, buttonStyle, icon, align, linkText, link, rel, referrerpolicy, srText } = blok;
 
   return (
     <SbEditable content={blok}>
@@ -14,6 +14,7 @@ export const SBCtaButton = ({ blok }) => {
         align={align}
         link={link}
         rel={rel}
+        referrerpolicy={referrerpolicy}
         srText={srText}
       >
         {linkText}


sherakama:
@shc314 do you have documentation on this requirement?

> To be used with the Project Muse button

I am asking because I haven't seen this type of requirement before and find it irregular.

William-Misiaszek:
Great work @shc314! 🤩

We are eager to have this merge checked and pushed into Production for SAA. End-Users currently do not have access to Project Muse without this fix (they should have it as part of their Alumni Membership).

@yvonnetangsu

src/components/cta/SAALinkButton.jsx
```diff
@@ -36,6 +37,7 @@ export const SAALinkButton = React.forwardRef(
       children,
       link,
       rel,
+      referrerpolicy,
```
sherakama, line 40: Instead of adding referrerpolicy as its own prop, can you pass it through the attributes prop please?

sherakama:
Just checking in on this, we have set up the referrer policy for the site here: https://github.com/SU-SWS/saa_alumni/blob/dev/netlify.toml#L19

Looking at the documentation in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#integration_with_html it seems that the referrer on the link would 'win' or be the higher priority but I've not confirmed that. 

As far as the two settings go, 

strict-origin-when-cross-origin vs no-referrer-when-downgrade

Strict origin encompasses what the downgrade option does, but also limits the information being sent (no query string or path). Does the Muse system need to know these attributes?

shc314:
@sherakama I've updated the Associated Issues section with all the information that I have. @William-Misiaszek would know more.

William-Misiaszek:
I believe it needs to have `no-referrer-when-downgrade`. In particular, the HTML code we want in the end is like the attached screenshot. This approach has been vetted on our end!
End Goal for CTA Button Update

src/components/cta/SAALinkButton.jsx
shc314: Do you mean as a custom attribute for the link? That will change the custom attributes for all CTA Button Links.

yvonnetangsu: @shc314 I'm no expert on referrer policy so I'll leave that part to @sherakama and @William-Misiaszek.

As for passing custom attributes to links in general, I believe Storyblok has added a new feature since we originally built the SAA Button Link component, which allows you to add any custom attributes to the Storyblok Link field through their UX, so for example:
Screenshot 2024-02-08 at 10 04 47 AM

So instead of adding an additional `referrerPolicy` field in Storyblok, I think what @sherakama meant is that you could allow the user to enter any additional attributes for the link through the "custom attributes" option in the Link field, and then have those attributes passed into the component.

sherakama:
> Hi @sherakama & @shc314
> 
> I believe it needs to have `no-referrer-when-downgrade`.

@William-Misiaszek, Do you have any supporting documentation, a ticket, a KB article, or something about this requirement? I am struggling to understand this requirement as the referrer header is generally considered [untrustworthy and unreliable](https://stackoverflow.com/questions/6023941/how-reliable-is-http-referer) and isn't recommended as something to build functionality by.

I apologize if I seem stonewalled on this but we've taken the time to set our header policies through a privacy and security lens and are meeting an ['A' grade](https://securityheaders.com/?q=alumni.stanford.edu&followRedirects=on). When something deviates from the established work it is my job to be curious and critical.

src/components/cta/SAALinkButton.jsx
sherakama: Oh, that would be nice. 

I was actually just referring to passing along the referrer propr through attributes in the template and saving all the conditional checks for rel, attributes, and referrer as the attributes prop is for adding things like the referrerpolicy attribute. 

@shc314 could you look into supporting the new custom attributes option in Storyblok as that would be a win for the whole codebase and site.

shc314: Thanks for your response, @yvonnetangsu . At first I thought it added the same custom attribute to all the links because after I added referrerpolicy to one of them, they all showed "1 attributes" but when I expanded it I saw it's just the _uid.

When I try to access referrerpolicy through attributes, I keep getting the error message that "attributes is not defined." Do you know how to access the custom attributes in the code?

shc314: @sherakama Sure. Does that need to be done without removing the rel prop, so as to not break it where it's currently in use?

When I tried to check what's in attributes, I got the error message that "attributes is not defined." Do you know what's going on?

sherakama: I set up a quick test to see where the attributes come through. 

Looks like they get added to the link directly as key -> value.

Screenshot 2024-02-09 at 09 53 46

URL: `/test/link-attributes-test`

You'll have to modify the sbLink component and pull the referrerpolicy key out of `props.link` and pass it to the actual `<Link>` through the `otherAttributes` var.

shc314:
@sherakama My code is ready for another review.

William-Misiaszek:
> > Hi @sherakama & @shc314
> > I believe it needs to have `no-referrer-when-downgrade`.
> 
> @William-Misiaszek, Do you have any supporting documentation, a ticket, a KB article, or something about this requirement? I am struggling to understand this requirement as the referrer header is generally considered [untrustworthy and unreliable](https://stackoverflow.com/questions/6023941/how-reliable-is-http-referer) and isn't recommended as something to build functionality by.
> 
> I apologize if I seem stonewalled on this but we've taken the time to set our header policies through a privacy and security lens and are meeting an ['A' grade](https://securityheaders.com/?q=alumni.stanford.edu&followRedirects=on). When something deviates from the established work it is my job to be curious and critical.

Hello @sherakama ,

Unfortunately when I try to insert referrerpolicy="strict-origin-when-cross-origin" the Project Muse system does not give SAA users access. Edith from Project Muse mentioned the referrerpolicy is necessary for their site to log-in users correctly.

Screenshot 2024-02-20 at 11 40 18

src/components/cta/SAALinkButton.jsx
sherakama: Yeah, Storyblok doesn't make it easy by mixing these custom attributes with the others.

I've updated the code a bit to parse the attributes on props.link and have added an allow-list. I did not include rel or target as that is already supported elsewhere, however, referrerpolicy and other in the list.

W3C does define which attributes are allowed on the anchor tag
https://www.w3schools.com/tags/tag_a.asp

src/utilities/sbLink.js
```diff
@@ -6,6 +6,25 @@ import { config } from './config';
...
+  /^tabindex$/,
+  /^role$/,
+  /^title$/,
+  /^lang$/,
+  /^inputmode$/,
+  /^data-.*/,
+  /^aria-.*/,
+  /^test-.*/,
+  /^id$/,
+];
```
sherakama, line 26: @yvonnetangsu & @shc314  how do you feel about this list of allowed attributes from Storyblok?

sherakama: I did not include ones that I felt should not be allowed as well as items like href, rel, and target as they are already supported another way.

https://www.w3schools.com/tags/tag_a.asp

sherakama:
> > > Hi @sherakama & @shc314
> > > I believe it needs to have `no-referrer-when-downgrade`.
> > 
> > 
> > @William-Misiaszek, Do you have any supporting documentation, a ticket, a KB article, or something about this requirement? I am struggling to understand this requirement as the referrer header is generally considered [untrustworthy and unreliable](https://stackoverflow.com/questions/6023941/how-reliable-is-http-referer) and isn't recommended as something to build functionality by.
> > I apologize if I seem stonewalled on this but we've taken the time to set our header policies through a privacy and security lens and are meeting an ['A' grade](https://securityheaders.com/?q=alumni.stanford.edu&followRedirects=on). When something deviates from the established work it is my job to be curious and critical.
> 
> Hello @sherakama ,
> 
> Unfortunately when I try to insert referrerpolicy="strict-origin-when-cross-origin" the Project Muse system does not give SAA users access. Edith from Project Muse mentioned the referrerpolicy is necessary for their site to log-in users correctly.
> 
> Screenshot 2024-02-20 at 11 40 18

@William-Misiaszek Thank you for following up. So this is how they lock the link to a site? They have you add the referring URL and enforce that header on their end? Alrighty then. Thank you for sharing that in this thread.

src/components/cta/SAALinkButton.jsx
sherakama: @shc314 can you re-test this to make sure it is showing the attribute as you'd expect it to? 
https://deploy-preview-805--stanford-alumni.netlify.app/shirley-test/

sherakama:
The Netlify checks are showing as failed. I manually rebuilt and you can see the preview site here: https://deploy-preview-805--stanford-alumni.netlify.app/shirley-test/

src/utilities/sbLink.js
sherakama: @yvonnetangsu @shc314

src/components/cta/SAALinkButton.jsx
shc314: @sherakama It works.

I understand your reason for leaving rel off the allow list. Since it is one of the default options under custom attributes, could it possibly confuse users who try to add it there?

sherakama:
Ok, moving forward with this. Thank you everyone for your input.

src/components/cta/SAALinkButton.jsx
sherakama: Aye, I agree that is not 'ideal'. I'd like to move our current 'rel' solution into the custom attributes but that would take some time and we'd need it prioritized. Let's move this forward and open up a new PR to test to see how/if it conflicts with the current solution.

sherakama:
I've added a follow up PR for the rel attribute: https://github.com/SU-SWS/saa_alumni/pull/809

William-Misiaszek:
Thank you @sherakama, @yvonnetangsu @shc314 ! 🎉

Our Membership Program Manager and her colleagues are thrilled 👏

PR #815, opened 2024-03-07
Task/HOM-1 update white hero banner gradient

shc314:
# READY FOR REVIEW

# Summary
- Changes text styles for the white hero banner gradient, to improve readability.

# Review By
- When possible

# Review Tasks

1. Check out this branch
2. Navigate to shirley-test page. Or create a page with a hero banner image with a White gradient, and enter text in the  Superheadline 2, Headline, and Subheadline fields. (Skip Superheadline 1 as the designer said they aren't using that now.)
3. Verify that:

- the font color for text on the white gradient is: #2E2D29
- Superheadline 2 and Subheadline both have text-shadow: rgba(255, 255, 255, .4) 0 0 10px, rgba(255, 255, 255, .6) 0 0 10px;
- Headline has text-shadow: rgba(255, 255, 255, .5) 0 0 12px;

@William-Misiaszek 
https://internal-saa.atlassian.net/browse/HOM-1


src/utilities/dataSource.js
shc314, line 1: This file is straight from dev. I made some changes locally but reverted them, and then Prettier wreaked havoc on that file, so I removed it and replaced it with the one on dev.

sherakama:
@yvonnetangsu would you mind giving this an eyeball? Thanks.

yvonnetangsu:
> @yvonnetangsu would you mind giving this an eyeball? Thanks.

Sure I'll take a look 😄

src/components/composite/hero.styles.js
```diff
@@ -0,0 +1,28 @@
+import { dcnb } from 'cnbuilder';
+
+export const serifSuper = ({ textColor }) =>
+  dcnb(
+    'su-block su-max-w-prose su-font-serif su-leading-display su-type-2 su-mx-auto su-mb-05em',
+    {
+      'su-text-shadow': textColor === 'su-text-white',
+      'su-text-shadow-white': textColor === 'su-black',
```
yvonnetangsu, line 8: @shc314 Wouldn't text color be `su-text-black` instead of `su-black` here? Or is that the value passed in from storyblok? Haven't pulled it down locally - just scanning the code at the moment 😄 When do you need this reviewed by? Ok if I want till later in the afternoon today?

shc314: Thank you for pointing that out. I wasn't aware of su-text-black, so I used su-black because it's also #2E2D29. I've just pushed the changes. Reviewing it this afternoon is fine.

yvonnetangsu: Hi @shc314 I believe `su-black` isn't a valid TW classname. It could still be rendering black text because it's inherited from the base TW styles (black text). Thanks for the update 😄 
https://tailwindcss.com/docs/text-color

yvonnetangsu (changes requested):
@shc314 I believe the "Scroll to explore" text would need to be black on the white gradient as well
https://deploy-preview-815--stanford-alumni.netlify.app/shirley-test/
Screenshot 2024-03-11 at 5 05 38 PM

yvonnetangsu:
Curious if this is ok for color contrast on mobile 🤔 Seems a bit hard to read even with the shadow. Might be something to ask SODA (office hour for a quick question perhaps?)
Screenshot 2024-03-11 at 5 07 44 PM

yvonnetangsu:
One option is to use su-text-black-true (#000) to push the contrast up a bit further, but even that might not be enough, so I'd suggest talking to SODA, and perhaps test with a few images. It is possible that you might need a more opaque white gradient on mobile since the font size is smaller.

src/tailwind/plugins/utilities/text-shadow/text-shadow.js
```diff
@@ -0,0 +1,18 @@
...
+          'rgba(255, 255, 255, 40%) 0 0 10px, rgba(255, 255, 255, 60%) 0 0 10px',
+      },
+      '.text-shadow-white-lg': {
+        textShadow: 'rgba(255, 255, 255, 50%) 0 0 12px',
+      },
+    };
+
+    addUtilities(newUtilities);
+  };
+};
```
yvonnetangsu, line 18: Thanks for adding custom TW utilities 👍🏼 😄

shc314: Thank you for your feedback @yvonnetangsu ! Your point about the text color contrast not being high enough on smaller screens is valid. I did try su-text-black-true but depending on the background image, it doesn't always have a contrast ratio of at least 4.5:1. I will bring this up with the designer.

I also realized that I forgot that the white gradient is different when the screen is smaller than xs, so I reinstated the white text for that size and made it black for larger sizes.

I'll let you know when my code is ready for review.

yvonnetangsu: Thank you for the updating and thinking about a11y @shc314! Now that we're on it, I suspect we need color contrast adjustments to the other variants also especially on mobile 🤔  You're right that a convo with the design team seems necessary 😄 

Screenshot 2024-03-15 at 10 46 53 AM

shc314: @yvonnetangsu My code is ready for review.
I tested the contrast for different background images at different breakpoints using TPGi Colour Contrast Analyser (CCA) and found some where it passes the minimum contrast ratio requirements. The SAA art director has agreed to check the contrast using CCA before using a background image and will only use those that pass.

William-Misiaszek: Hello @yvonnetangsu and @shc314 

Is this feature ready for production? We did get confirmation our design team will check contrast ratios with a CCA for the White Gradient. This color option is worth having for them!

When will the white gradient be available in production?

yvonnetangsu: @shc314 @William-Misiaszek I'll approve this, but please make sure the design team do pick a photo that will meet contrast ratio - especially on mobile breakpoints. For example, this image on the test page would not work for mobile:
Screenshot 2024-04-03 at 10 43 39 AM

yvonnetangsu: @shc314 I understand that you'd like to get this into production as soon as you can, but it is still worth using `su-text-black-true` when you're using the white gradient option - perhaps in another PR. Every little bit helps 😄

src/components/composite/hero.js
```diff
@@ -108,7 +112,12 @@ const Hero = ({
               </div>
             )}
             {!isHideScroll && (
-              <div className="su-text-center su-text-white su-grow-0 su-rs-mt-5 su-font-serif su-font-regular su-text-19 md:su-text-22">
+              <div
+                className={dcnb(
```
yvonnetangsu, line 116: @shc314 This is for future also - it would be nice to keep all the style code in hero.styles.js. We have a mix here, some in the hero.js and some in hero.styles.js. Would be nice to keep everything consistent.

yvonnetangsu (approved):
Approving but as discussed, please make sure design team selects images that meet contrast requirement for accessibility 😄 Please also consider keeping convention consistent.

yvonnetangsu:
It looks like @shc314 published the test page with new images. These do look better, but I believe only the one at the bottom  is actually fully compliant 🤔 . Leaving out the superheading also helps a bit since the higher part of the image doesn't have as much of the white gradient color. cc @William-Misiaszek 

Screenshot 2024-04-03 at 10 56 36 AM

yvonnetangsu:
I don't believe the cypress test has anything to do with this PR so I'm going to merge this into `dev` @sherakama

src/tailwind/plugins/utilities/text-shadow/text-shadow.js
shc314: Will do.

src/components/composite/hero.js
shc314: I didn't move it because it didn't have conditional CSS, but I can put it in hero.styles.js. Should I also put both hero.js and hero.styles.js in their own Hero directory?

yvonnetangsu: No worries @shc314 ! 😄  in our newer repos, we do put almost every component and their style js file in a separate directory to keep things a bit cleaner. If we have a style js file, we usually put all the TW classes and things into that file whether there is a conditional or not (_usually_ - since there were times when we missed things or when there were other vendors working on things as well).

sherakama:
> I don't believe the cypress test has anything to do with this PR so I'm going to merge this into `dev` @sherakama

Eric noticed it as well and there is [WIP](https://github.com/SU-SWS/saa_alumni/pull/824).

PR #820, opened 2024-03-13
DS-584: Protect Protected Content at the Edge

sherakama:
# READY FOR REVIEW

https://stanford.atlassian.net/browse/DS-584

https://my2.siteimprove.com/Accessibility/1088922/NextGen/Issue/1?ruleId=1&pageSegments=&issueKind=1&exceptTags=1,2&conformance=&siteTargetIssueKinds=1,2&wcagVersion=21&siteTarget.wcagVersion=21&siteTarget.conformanceLevels=0,1,3,4&siteTarget.issueKinds=1,2

# Summary
- Creates an edge function that looks for the auth cookie. 
- If no auth cookie redirect to login
- Add only fully auth required paths to netlify.toml
- The hybrid auth paths will have to stay as they are
- The actual authentication validation still happens on the functions
- This is to 'hide' the loading pages as they are flagged as inaccessible in Siteimprove

# Review By (Date)
- End of the sprint

# Criticality
- Normal, this is for the June 1st a11y improvements

# Review Tasks

1. Navigate to the [deploy preview](https://deploy-preview-820--stanford-alumni.netlify.app/)
2. Click around some and ensure pages still load as they are supposed to
3. As an unauthenticated user, navigate to the membership join page and ensure you are redirected to login
4. As an unauthenticated user, navigate to the custom journeys page and ensure you are redirected to login
5. As an unauthenticated user, navigate to a notify me page and ensure you are redirected to login
6. As an unauthenticated user, navigate to a reserve page and ensure you are redirected to login
7. When you log in, ensure you land on the page you were trying to get to
8. Review the code


netlify.toml
```diff
@@ -36,6 +36,26 @@
...
+  path = "/travel-study/custom-journeys/request"
+  function = "protected-content"
+
+[[edge_functions]]
+  path = "/travel-study/destinations/*/reserve"
+  function = "protected-content"
+
+[[edge_functions]]
+  path = "/travel-study/destinations/*/notify"
+  function = "protected-content"
```
sherakama, line 53: @rebeccahongsf I'm trying to remember if there was a third URL here that was fully behind authentication. Do you recall?

netlify/edge-functions/protected-content.js
```diff
@@ -0,0 +1,13 @@
...
+    // Pass through.
+    return;
+  }
+
+  const url = new URL('/api/auth/login', req.url);
+  const sourceURL = new URL(req.url);
+  url.searchParams.set('final_destination', sourceURL.pathname);
+  // eslint-disable-next-line consistent-return
+  return Response.redirect(url);
+};
```
ericbakenhus, line 13: @sherakama I'm still getting redirected after the page loads. For example, if I disable JS and try to visit a protected page I just land on a static page with a loading icon (the edge function never redirects me).

Screenshot 2024-03-13 at 6 10 07 PM

ericbakenhus: Yep, works fine. 😆

ericbakenhus: @sherakama Ok, now I've run into an _actual_ issue. When Stanford Pass sends me back to `/auth`, it bounces me to `/api/auth/callback` which then redirects to my final destination (`/membership/join`) which is then getting caught in the edge function again and redirecting back though the login process.

Here are those calls for example:

Screenshot 2024-03-13 at 6 30 20 PM

sherakama: 🤔 Yeah, that is happening to me and I swear it wasn't before. Who dropped the cookie?

sherakama: Ok, I'm seeing the correct set-cookie headers on the way out from the auth function and in the browser but no headers or cookies on the edge function. Why?

Screenshot 2024-03-14 at 09 12 50
Screenshot 2024-03-14 at 09 10 07
Screenshot 2024-03-14 at 09 13 38

sherakama: Bug? https://issues.chromium.org/issues/40288265

sherakama: Hmm. `SameSite=Strict` on a 302 seems to not be passed...

ericbakenhus: I'm repeating unverified internet knowledge, but it sounds like the entire redirect chain is considered with `SameSite=Strict` and if there's a site in that chain that isn't in the same site, the cookie doesn't get sent. The chain includes `https://myaccount-sso-uat.stanford.edu/api/sso/auth`. Maybe that's why it's not being sent when it eventually gets redirected back to the final destination page? 

I don't suppose there's an easy way to use `SameSite=Lax` on the initial auth then upgrade it to `SameSite=Strict` after all the redirects?

ericbakenhus: Maybe switch to the `*-session` cookie to lax and check for that one instead?

sherakama: 🤔 Interesting
Yeah, I am leaning toward opening up the session cookie. Gonna try changing the status code to see if that does anything first.

sherakama: The question I have is, why was it working before?

sherakama: ... Because the website makes an XHR _AFTER_ the user lands on the page and the cookie is set.

sherakama:
Interestingly enough someone is suggesting a meta refresh to get around this: https://stackoverflow.com/questions/42216700/how-can-i-redirect-after-oauth2-with-samesite-strict-and-still-get-my-cookies

ericbakenhus:
> Interestingly enough someone is suggesting a meta refresh to get around this: https://stackoverflow.com/questions/42216700/how-can-i-redirect-after-oauth2-with-samesite-strict-and-still-get-my-cookies

One sec...gotta check which decade we're in. 😄 

But seriously, we could do something like that. Have `/auth` be a real page and trigger the redirect on the client. That's basically what the Next.js apps are doing.

sherakama:
> But seriously, we could do something like that. Have the ~`/auth`~ `/callback`(?) page be a real page and trigger the redirect on the client. That's basically what the Next.js apps are doing.

The LAMBDA could just spit back the HTML as well... But yeah, I'll stay in this century and live with LAX.

netlify.toml
sherakama: @rebeccahongsf ping

doc/adr/0002-edge-protection-for-protected-pages.md
```diff
@@ -0,0 +1,39 @@
+# 2. Edge protection for protected pages
```
sherakama, line 1: @ericbakenhus & @rebeccahongsf 

Threw this in here as well.

PR #842, opened 2024-04-27
DS-654 | @ericbakenhus | Hide empty uls

ericbakenhus:
# READY FOR REVIEW

# Summary
- Hides a couple `ul`s if they're empty.

# Review By (Date)
- Whenever

# Criticality
- Low

# Review Tasks

## Setup tasks and/or behavior to test

1. Check out this branch or use the [deploy preview](https://deploy-preview-842--stanford-alumni.netlify.app/)
2. Navigate to [/groups/identities/](https://deploy-preview-842--stanford-alumni.netlify.app/groups/identities/)
3. Verify there aren't a bunch of empty `ul`s below "Shared Identity" (you can use an a11y tool to check for 1.3.1 issues)

For example (dev site on left, deploy preview on right):
Screenshot 2024-04-29 at 10 39 17 AM

4. Look at code for the associate list (since we can't really test it easily) and make sure the [page itself](https://deploy-preview-842--stanford-alumni.netlify.app/stanford-associates/directory/) still functions correctly

# Associated Issues and/or People
- DS-654

PR #843, opened 2024-04-29
Release v4.3.17

ericbakenhus:
Release v4.3.17

Includes:
- Hiding empty ULs: #842


ericbakenhus:
The Travel Study test is also having issues since the refresh. Something to look into.

PR #846, opened 2024-05-21
DS-469: Update print version of digital membership card for SAA and GSB

stfurkan:
# READY FOR REVIEW

# Summary
- DS-469: Update print version of digital membership card for SAA and GSB

# Review By (Date)
- End of sprint

# Criticality
- 3

# Review Tasks

1. Check out this branch
2. Navigate to `/membership/saacard/` and `/membership/gsbcard/`
3. Make sure that printed card shows the card color correctly (without updating any additional setting in the browser's print configuration)


ericbakenhus (approved):
Hmm, Firefox and Chrome seem like they might be ignoring some of those existing print-only positioning rules on the print preview. Something that's actually noticeable now that the card is visible. 😅 Might be something to check out in the future.

Everything LGTM! 🎉

PR #855, opened 2024-06-05
Nojira | @ericbakenhus | test background config

ericbakenhus:
# Just Testing


ericbakenhus:
@sherakama After some testing I've found when setting the function URL for background function:
- The UI will shows the prod URL for all deploy previews
- If you start the URL with `/api` it won't work (I'm assuming it conflicts with Netlify's handling of Gatsby functions)
- Using a URL that doesn't start with `/api` works

PR #858, opened 2024-06-17
DS-690 | @ericbakenhus | Scheduled import function

ericbakenhus:
# NOT READY FOR REVIEW

# Summary
- DS-690: Add scheduled and background functions to handle syncing events from Google Sheets to Storyblok

# Review By (Date)
- End of sprint

# Criticality
- Low

# Review Tasks

1. Check out this branch
2. Navigate to...
3. Verify...

- DS-690

### DIFF
--- a/netlify/functions/event-import-background.mts
+++ b/netlify/functions/event-import-background.mts
@@ -1,0 +1,73 @@
+import dotenv from 'dotenv';
+import { Context, type Config } from '@netlify/functions';
+import { GoogleSpreadsheet, type GoogleSpreadsheetWorksheet } from 'google-spreadsheet';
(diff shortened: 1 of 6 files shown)


ericbakenhus:
Thinking about rebasing this on dev so we can properly test imports before any public launch... 🤔

ericbakenhus:
> > Thinking about rebasing this on dev so we can properly test imports before any public launch... 🤔
> 
> I like the idea of having the data flow into storyblok ahead of pushing anything public.

Coolios. I'll do that then. Thanks! 👍

ericbakenhus:
Replaced with #867.

PR #860, opened 2024-06-18
DS-687 | @ericbakenhus | Add generated fields for Algolia config

ericbakenhus:
# READY FOR REVIEW

# Summary
- Add fields needed for Algolia config

# Review By (Date)
- When possible

# Criticality
- High

# Review Tasks

1. Look at code

- DS-687

### DIFF
--- a/src/utilities/synchronizedEvents.js
+++ b/src/utilities/synchronizedEvents.js
@@ -1,3 +1,7 @@
+import { render } from 'storyblok-rich-text-react-renderer-ts';
+import sanitizeHtml from 'sanitize-html';
+import { luxonDate } from './dates';
+
 const isString = (val) => typeof val === 'string';
 
 const isLink = (val) =>
@@ -54,4 +58,30 @@ export const mergeEventOverrides = (eventContent) => {
   });
 
   return merged;
+};
+
+export const storyToAlgoliaEvent = (story) => {
+  const storyId = story.data.story.uuid;
+  const eventData = story.data.story.content;
+  const mergedEventData = mergeEventOverrides(eventData);
+  const startTimestamp = mergedEventData.start
+    ? luxonDate(mergedEventData.start).toUnixInteger()
+    : null;
+  const endTimestamp = mergedEventData.end
+    ? luxonDate(mergedEventData.end).toUnixInteger()
+    : null;
+  const descriptionText = mergedEventData.description
+    ? sanitizeHtml(render(mergedEventData.description), {
+        allowedTags: [],
+        allowedAttributes: {},
+      })
+    : null;
+
+  return {
+    objectID: storyId,
+    startTimestamp,
+    endTimestamp,
+    descriptionText,
+    ...mergedEventData,
+  };
 };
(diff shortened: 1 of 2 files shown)


src/utilities/synchronizedEvents.js
```diff
@@ -55,3 +59,29 @@ export const mergeEventOverrides = (eventContent) => {
...
+export const storyToAlgoliaEvent = (story) => {
+  const storyId = story.data.story.uuid;
+  const eventData = story.data.story.content;
+  const mergedEventData = mergeEventOverrides(eventData);
+  const startTimestamp = mergedEventData.start
+    ? luxonDate(mergedEventData.start).toUnixInteger()
+    : null;
+  const endTimestamp = mergedEventData.end
+    ? luxonDate(mergedEventData.end).toUnixInteger()
+    : null;
```
ericbakenhus, line 72: Unix timestamps based on merged event datetimes.

src/utilities/synchronizedEvents.js
```diff
@@ -55,3 +59,29 @@ export const mergeEventOverrides = (eventContent) => {
...
+  const descriptionText = mergedEventData.description
+    ? sanitizeHtml(render(mergedEventData.description), {
+        allowedTags: [],
+        allowedAttributes: {},
+      })
+    : null;
```
ericbakenhus, line 78: Text version of description, if we ever need search for that content.

stfurkan (approved):
LGTM! Thanks @ericbakenhus :)

ericbakenhus:
@stfurkan Can you think of any other fields we might need to generate for searches, facets, etc?

Also, it looks like DS-687 includes the Algolia config for facets, but it seems like it would make more sense for you to handle those as needed while building the UI. Would you prefer to handle those?

I'm going to take a quick pass at minimally configuring what I think makes sense in Algolia, but please feel free to change/add/edit anything in there if it's not what you need (especially if I mess something up 😆).

stfurkan:
> @stfurkan Can you think of any other fields we might need to generate for searches, facets, etc?
> 
> Also, it looks like [DS-687](https://stanford.atlassian.net/browse/DS-687) includes the Algolia config for facets, but it seems like it would make more sense for you to handle those as needed while building the UI. Would you prefer to handle those?
> 
> I'm going to take a quick pass at minimally configuring what I think makes sense in Algolia, but please feel free to change/add/edit anything in there if it's not what you need (especially if I mess something up 😆).

I haven't tried the lat/long data but we might need an update for it, if it's not using the correct type. Sure, I can try to config facets while I'm building the UI (I added a couple of facets to the main dev index). I'll ask Shea or you, if I have any questions :) Thanks.

ericbakenhus:
> I haven't tried the lat/long data but we might need an update for it, if it's not using the correct type. 

Oh yeah! I can swap those to be numbers and generate what Algolia wants for geolocation data.

src/utilities/synchronizedEvents.js
```diff
@@ -0,0 +1,96 @@
...
+  const geo =
+    mergedEventData.latitude && mergedEventData.longitude
+      ? {
+          lat: mergedEventData.latitude,
+          lng: mergedEventData.longitude,
+        }
+      : null;
```
ericbakenhus, line 86: Add geolocation data formatted for Algolia.

PR #863, opened 2024-06-19
DS-687 | @ericbakenhus | Remove description text

ericbakenhus:
# READY FOR REVIEW

# Summary
- Removes generated description text

# Review By (Date)
- Retro

# Criticality
- High

# Review Tasks

1. Look at code

- DS-687

### DIFF
--- a/src/utilities/synchronizedEvents.js
+++ b/src/utilities/synchronizedEvents.js
@@ -1,5 +1,3 @@
-import { render } from 'storyblok-rich-text-react-renderer-ts';
-import sanitizeHtml from 'sanitize-html';
 import { luxonDate } from './dates';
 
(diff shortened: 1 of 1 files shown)


ericbakenhus:
Description isn't shown on the card currently, so we probably won't want to include it in the search (and I don't feel like troubleshooting it). So, I'm jettisoning it.

sherakama:
What were you having trouble with?

ericbakenhus:
> What were you having trouble with?

The HTML sanitizer (config'd to strip out all tags and attributes to leave only text) was throwing a nondescript error on the render functions output. If it's later decided to include the description in some way I'll properly troubleshoot and add this back in.

PR #867, opened 2024-06-20
DS-690, DS-691 | @ericbakenhus | Scheduled function

ericbakenhus:
# READY FOR REVIEW

# Summary
- Add scheduled and background functions to handle event import from Google Sheets
- Tweak deploy function to better handle functions and generate correct regions and tags

# Review By (Date)
- 🤷 

# Criticality
- Low

# Review Tasks

1. (optional) Delete some/all events currently in the Dev space
2. Open the logs (background functions are real-time only 😐)
3. Trigger one or more syncs with the Import Synced Events task
4. See events being imported and whatnot 

- DS-690
- DS-691


package.json
```diff
@@ -28,7 +28,7 @@
     "gatsby-plugin-netlify": "^4.4.0",
     "gatsby-plugin-postcss": "^5.25.0",
     "gatsby-plugin-react-helmet": "^5.25.0",
-    "gatsby-plugin-robots-txt": "^1.7.0",
+    "gatsby-plugin-robots-txt": "~1.7.0",
```
ericbakenhus, line 31: 1.8 _requires_ Gatsby 5.

netlify/functions/trigger-event-import.mts
```diff
@@ -0,0 +1,42 @@
...
+    console.error('Error duing request: ', err);
+  }
+
+  console.log('=== END Trigger Event Import ===');
+  return Response.json({ statusCode: 200 });
+};
+
+export const config: Config = {
+  schedule: '0 15 * * *', // Every day at 3 PM UTC (8 AM PT)
+};
```
ericbakenhus, line 42: New scheduled event. Triggers the import background function once a day at 8AM PT.

netlify/functions/event-import-background.mts
ericbakenhus, line 1: New import background function.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,287 @@
...
+    // stop, test, or run
+    const mode = process.env.EVENT_IMPORT_MODE ?? 'stop';
+
+    if (mode === 'stop') {
+      throw new Error('Import mode set to "stop"');
+    }
+
+    console.log(`Running in ${mode} mode`);
+
+    const run = mode === 'run';
```
ericbakenhus, line 34: Adding a run mode env var. `stop` prevents it from running. `test` runs the function but doesn't trigger any data-changing requests (i.e. performs a dry run). `run` fully runs everything.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,287 @@
...
+    console.log('Fetching Cvent data...');
+    const cventDoc = new GoogleSpreadsheet(cventSheetId, auth);
+    await cventDoc.loadInfo();
+    const cventSheet = cventDoc.sheetsByIndex[0];
+    const cventRows = await cventSheet.getRows();
+    console.log('Fetching Cvent data done!');
+    
+    cventRows.forEach((row) => {
+      googleStories.push(googleRowToStory(row.toObject(), 'Cvent'));
+    });
```
ericbakenhus, line 77: Grab everything in the two source Google sheets, process them through `googleRowToStory` to convert them into valid story objects for post/put.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,370 @@
...
+    console.log('Fetching Cvent data...');
+    const cventDoc = new GoogleSpreadsheet(cventSheetId, auth);
+    await cventDoc.loadInfo();
+    const cventSheet = cventDoc.sheetsByIndex[0];
+    const cventRows = await cventSheet.getRows();
+    console.log('Fetching Cvent data done!');
+    
+    cventRows.forEach((row) => {
+      googleStories.push(googleRowToStory(row.toObject(), 'cvent'));
+    });
```
ericbakenhus, line 82: Grab everything in the two source Google sheets, process them through `googleRowToStory` to convert them into valid story objects for posting/putting to SB.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,370 @@
...
+      filter_query: { __or: [
+        { end: { lt_date: archiveCutoff }},
+        { endOverride: { lt_date: archiveCutoff }}
+      ]}, 
+      content_type: 'synchronizedEvent', 
+      version: 'draft' 
+    }) ?? [];
+    const sbEvents = [...sbPublishedEvents?.data?.stories?.map((s) => ({ ...s, isPublished: true })), ...sbUnpublishedEvents?.data?.stories?.map((s) => ({ ...s, isPublished: false }))];
+    const oldArchivedEvents = [...oldArchivedPublishedEvents?.data?.stories, ...oldArchivedUnpublishedEvents?.data?.stories].filter((s) => !!s);
+    console.log('Fetching Storyblok events done!');
```
ericbakenhus, line 134: Grab `everything` we need from SB (our relevant datasources, draft and published events in `events/sync`, and draft and published events in `events/sync-archive`).

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,370 @@
...
+        const value = existing
+          ? { ...existing, storyblok: story }
+          : { google: undefined, storyblok: story }
+
+        syncedEvents.set(id, value);
+      } else {
+        console.log('No ID for SB event: ', story.full_slug);
+        manualEvents.set(story.id, story);
+      }
+    });
```
ericbakenhus, line 180: Here we are creating maps to link up any events (based on externalId) that are in both Google and SB. It also allows us to see any that _aren't_ in both. As a bonus we're grabbing all the incoming formats and tags that are coming from Google sheets for later.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,370 @@
...
+          datasource_entry: {
+            name: tag,
+            value: tag,
+            datasource_id: identityTagsDatasourceId,
+          }
+        } as any);
+      } catch (err) {
+        console.log('Error during identity tags datasource update: ', err);
+      }
+    }
```
ericbakenhus, line 229: Create sets (to avoid duplicates) of the difference between existing and incoming datasources. Once we have that we update the relevant datasources in SB with new values.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,370 @@
...
+            }
+            console.log('Unpublished!');
+          }
+        }
+      } catch(err) {
+        console.error('Error during proccessing: ', err);
+      }
+
+      console.log('Processing complete: ', id);
+    }
```
ericbakenhus, line 314: Process all events in Google and/or SB.

- If an event exists in both
    - Check if there are any changes
    - If there are changes, merge the new data into the old data and update the event in SB
- If an event only exists in Google
    - Post it to SB
- If an event only exists in SB
    - Unpublish it
    - If it's "old" move it to the archive folder

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,371 @@
...
+            });
+          }
+          console.log('Done!');
+        }
+      } catch (err) {
+        console.error('Error during proccessing: ', err);
+      }
+
+      console.log('Processing complete: ', id);
+    }
```
ericbakenhus, line 345: Run through all the manual events (no external ID). If it's old, unpublish it and move it to the archive.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,371 @@
...
+        if (run) {
+          await storyblokManagement.delete(`spaces/${spaceId}/stories/${story.id}`, {});
+        }
+        console.log('Deleted!');
+      } catch (err) {
+        console.error('Error during processing: ', err);
+      }
+
+      console.log('Processing complete: ', story.id);
+    }
```
ericbakenhus, line 361: Run through and delete "old" archived events (end date older than 180 days).

netlify/functions/handle-deploy.mts
```diff
@@ -0,0 +1,164 @@
...
+    if ((isFolder || !isEvent) && !isEventFolder) {
+      // Trigger rebuild and stop
+      console.log('(Un)publish event detected. Deploying...');
+      if (run) {
+        await fetch(deployUrl, { method: 'POST' });
+      }
+      console.log('Deploy triggered');
+      console.log('=== END Deploy Background Function ===');
+      return new Response('Accepted', { status: 202 });
+    }
```
ericbakenhus, line 88: Checking if it's a folder now as well (triggers a deploy).

netlify/functions/handle-deploy.mts
```diff
@@ -0,0 +1,164 @@
...
+      });
+
+      storiesToProcess = storiesRes?.data?.stories;
+
+      console.log('Event folder (un)publish event detected. Deploying...');
+      if (run) {
+        await fetch(deployUrl, { method: 'POST' });
+      }
+      console.log('Deploy triggered');
+    }
```
ericbakenhus, line 121: If the event or event/sync folder is (un)published, we grab all the stories in event/sync and (un)publish them.

netlify/functions/handle-deploy.mts
```diff
@@ -0,0 +1,164 @@
...
+
+    storiesToProcess = storiesToProcess.filter((story) => story.full_slug.startsWith('events/sync/'));
+
+    const regions = await storyblokContent.get('cdn/datasource_entries', {
+      datasource: 'synchronized-event-regions',
+      dimension: 'us-or-international',
+    });
```
ericbakenhus, line 128: Also grabbing the regions datasource so we can categories region as US or international.

ericbakenhus:
Honestly, it would be nice to break out a lot of this and add proper functional tests for each part individually. Maybe we can leave that for future us...

ericbakenhus:
New source fixes just appeared. Once sec...

netlify/functions/event-import-background.mts
sherakama: Could you throw this comment into the code or an ADR for future us.

sherakama:
This is a big lift. Thanks. This will take me a beat to go through.

ericbakenhus:
> This is a big lift. Thanks. This will take me a beat to go through.

@sherakama You might want to hold off for a while. I've had to continue tweaking things and bolting on additional functionality as the source data becomes clearer. It needs some clean up.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,383 @@
...
+import { compareStoryContent, googleRowToStory, combineStories, setStoryRegion } from '../../src/utilities/synchronizedEvents';
+import { luxonDate } from '../../src/utilities/dates';
+import { DateTime } from 'luxon';
+
+dotenv.config();
+
+// Keep SB API happy
+const delay = (m = 250) => new Promise((resolve) => {
+  setTimeout(() => resolve(null), m);
+});
```
sherakama, line 15: Rate limit?

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,383 @@
...
+        'https://www.googleapis.com/auth/spreadsheets',
+      ],
+    });
+    console.log('Google auth done!');
+
+    const googleStories: any[] = [];
+
+    console.log('Fetching Hiverbrite data...');
+    const hivebriteDoc = new GoogleSpreadsheet(hivebriteSheetId, auth);
+    await hivebriteDoc.loadInfo();
```
sherakama, line 69: I imagine this throws an error if it fails to load for whatever reason. For future debugging us, this would get logged (as much as it does) and then the try again 2 more times would happen.

netlify/functions/event-import-background.mts
sherakama: There is a `rateLimit` param that you can pass the Storyblok client config.

sherakama: https://github.com/storyblok/storyblok-js-client#class-storyblok

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,383 @@
...
+            datasource_entry: {
+              name: format,
+              value: format,
+              datasource_id: formatDatasourceId,
+            }
+          } as any);
+          await delay();
+        }
+      } catch (err) {
+        console.log('Error during format datasource update: ', err);
```
sherakama, line 207: So if a datasource entry fails to save, the story entry would still get the value, however, it would be orphaned data and just not show up in the json result of the API data fetch for that story later?

netlify/functions/event-import-background.mts
ericbakenhus: Well even with their default rate limits in place and no parallelization, I still managed to hit a rate limit. Considering the number of bugs and jank I ran into, I don't know that I fully trust it.

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,383 @@
...
+            console.log('Unpublished!');
+          }
+
+          if (isOld) {
+            console.log('Story is old. Moving...');
+            if (run) {
+              await storyblokManagement.put(`spaces/${spaceId}/stories/${storyblok.id}`, {
+                story: { 
+                  ...storyblok,
+                  full_slug: `events/sync/archived/${storyblok.slug}`,
```
sherakama, line 310: Do we really care if we keep everything year after year? Me thinks this is fine and a future us problem.

sherakama: On build, we probably want to filter out the archived path...

netlify/functions/event-import-background.mts
```diff
@@ -0,0 +1,383 @@
...
+          console.log('Done!');
+        }
+      } catch (err) {
+        console.error('Error during proccessing: ', err);
+      }
+
+      console.log('Processing complete: ', id);
+    }
+
+    for (const story of oldArchivedEvents) {
```
sherakama, line 358: Ah, ignore previous comments about keeping things.

netlify/functions/event-import-background.mts
ericbakenhus: We fully delete archived events that are 180 days old. 

Do you mean keep it out of the build/sitemap? Yeah, should probably do that. It used to be in the sync folder, but got moved out to simplify things.

sherakama (approved):
Nice work. Looks good to me. Thanks for the effort and trudging through some SB wackiness and data-handling silliness.

netlify/functions/event-import-background.mts
sherakama: > Considering the number of bugs and jank I ran into, I don't know that I fully trust it

Almost got a spit take out of me today too!

netlify/functions/event-import-background.mts
sherakama: Ah, yeah, out of site map, federated search, and the build would be good.

sherakama: Being unpublished should generally do those things but some extra attention would be a belt + suspenders.

netlify/functions/event-import-background.mts
ericbakenhus: I _think_ only scheduled functions retry 3 times if they don't return the expected response. I can check though.

It should log the error under relevant "Fetching..." line. Not that it gets saved anywhere because background functions aren't "real" functions.

Maybe we can toss in a slack ping if things go wrong?

ericbakenhus: I _could_ add manual retries to each individual fetch through... 🤔

netlify/functions/event-import-background.mts
ericbakenhus: I can't remember anyone, but I want to say orphaned values get saved and passed through. They just don't show up in the editing UI. I could be wrong about that though.

Either way, yes that's basically true. Went with a datasource so we could lock down the available options in the UI while still allowing them to (potentially) manually edit the values if needed. Ideally this is used less and less over time as we add all the potential values. It's also self healing and should correct any missed values on future runs.

netlify/functions/event-import-background.mts
ericbakenhus: Ah, but I apparently forgot to fix this URL. Well then.

ericbakenhus: The archive folder is excluded from the sitemap now. Build is based on content type, so that should already be good. Double checked that the crawler is excluding the entire events folder. Should be good to go. 👍

netlify/functions/event-import-background.mts
sherakama: Yeah, I like this pattern overall. One thing I felt was strange that SB did was to allow orphaned values to continue to exist, but it does allow for some resilience and I've come to like and expect it.

netlify/functions/event-import-background.mts
sherakama: Not so much worried about the re-tries over having some visibility to when and how often things go wrong so that we may take action or troubleshoot.

sherakama: Not a today thing however.

ericbakenhus: Finally looked into it. If there's an unhandled exception in a background function it will retry 2 more times. We're catching almost everything currently, so we probably won't run into that behavior without something going catastrophically wrong.

sherakama: Ty

PR #869, opened 2024-06-24
DS-766 | @ericbakenhus | Magazine redirects

ericbakenhus:
# READY FOR REVIEW

# Summary
- Adds some magazine redirects

# Review By (Date)
- End of sprint (or before 7/12)

# Criticality
- Low

# Review Tasks

1. Look at redirects
2. Make sure they redirect on the deploy preview

- DS-766


sherakama:
I see you've run into Netlify's lack of documentation for this format
If you're looking for examples try this out for size https://github.com/SU-SWS/ood_online_giving/blob/dev/static/_redirects

sherakama:
Redirects will execute on the first match so put your most specific redirects at the top of the document.

ericbakenhus:
> Redirects will execute on the first match so put your most specific redirects at the top of the document.

They gave me two redirects with the same source URL, so I'm just waiting for a fix there. Everything else should be good to go. 👍

sherakama (approved):
GTG after redirect for article_id 88299 is sorted out.

PR #886, opened 2024-07-18
DS-793 | @ericbakenhus | Allow switching timezones on events

ericbakenhus:
# READY FOR REVIEW

# Summary
- Adds select to switch time zones for each event

# Review By (Date)
- Retro

# Criticality
- Low

# Review Tasks

1. Check out this branch
3. Verify that the time zone on each event is a dropdown _if the event's time zone is different than your system's time zone_
4. Verfiy that the time zone on each event is simple text _if the event's time zone is the same as your system's time zone_
5. Verify that changing an event's time zone changes the start/end datetimes to match

- DS-793


sherakama (commented):
k

PR #889, opened 2024-07-19
DS-793 | @ericbakenhus | Handle when different timezones with the same offset

ericbakenhus:
# READY FOR REVIEW

# Summary
- Handle the edge case where timezones are different but the offsets are the same (thanks Arizona)

# Review By (Date)
- Retro

# Criticality
- Low

# Review Tasks

1. Check out this branch
3. Create/edit an event so that the timezone is `America/Phoenix`
4. Verify you see something like this for that event (assuming you're in PDT currently):

Screenshot 2024-07-19 at 6 51 59 PM

- DS-793


sherakama (commented):
https://www.change.org/p/eliminate-daylight-savings-time

PR #904, opened 2024-07-30
DS-716 | @ericbakenhus | Mobile filters

ericbakenhus:
# READY FOR REVIEW

# Summary
- Adds mobile filters
- Adds mobile search bar

# Review By (Date)
- Soon/Retro

# Criticality
- High

# Review Tasks

1. Check out this branch
3. Verify that on smaller screen sizes, the search bar acts as a button and a filter button appears
4. Verify that both buttons open a side drawer
5. Verify that all the filters and the search bar are available in the slide drawers and working as expected

NOTE: I removed the "Next X days" radio options from the mobile date filter as it seems a bit much on mobile. I will offer to add some "Next X days" preset buttons if the client wants them after review.

- DS-716


sherakama:
I'll have a look.

sherakama:
Overall looking really good.

A couple of items and nitpicky details.

* Date options are looking really small IMHO
* Clear filters is hidden under options
* Console error about switching to controlled vs uncontrolled on date filter
* Location mobile city search isn't working. 
  * Missing `refine` error 
* Location city and sub-filter options should have their drop-down changed to an inline element instead of an absolute element as it overflows 
* Sub-facets 'view results' button closes the first panel but not the second panel as well

Screenshot 2024-07-30 at 17 10 34
Screenshot 2024-07-30 at 17 10 46
Screenshot 2024-07-30 at 17 11 01
Screenshot 2024-07-30 at 17 11 14
Screenshot 2024-07-30 at 17 11 35
Screenshot 2024-07-30 at 17 11 46
Screenshot 2024-07-30 at 17 12 03
Screenshot 2024-07-30 at 17 12 41
Screenshot 2024-07-30 at 17 13 13
Screenshot 2024-07-30 at 17 13 27
Screenshot 2024-07-30 at 17 14 38
Screenshot 2024-07-30 at 17 14 59
Screenshot 2024-07-30 at 17 16 00
Screenshot 2024-07-30 at 17 16 23
Screenshot 2024-07-30 at 17 16 48

src/components/events-discovery/components/Filters/MobileFilterFooter.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { dcnb } from 'cnbuilder';
+import { useClearRefinements } from 'react-instantsearch';
+
+export const MobileFilterFooter = ({
+  onCloseMenu,
+  className,
+  showClear = true,
+}) => {
+  const { canRefine, refine: clearRefinements } = useClearRefinements();
```
sherakama, line 10: You'll likely need to add the city's `useRadialGeoSearch` here in this check as well as it isn't a facet and not included in the refinementlist.

src/components/events-discovery/components/LocationFilter/MobileLocationFilter.jsx
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @type {React.FC<Props>}
+ * @returns {React.ReactElement}
+ */
+export const MobileLocationFilter = ({ onCloseMenu }) => {
+  const { items } = useCurrentRefinements({
+    includedAttributes: ['usRegion', 'intRegion', 'state', 'country'],
+  });
+
+  const count = useMemo(
```
sherakama, line 25: If you want to include city in this count you'll need to grab and check for a `name` value in `useRadialGeoSearch`

src/components/events-discovery/eventsDiscovery.jsx
```diff
@@ -88,33 +96,22 @@ const EventsDiscovery = () => (
...
-          // Location Search params.
-          if (
-            indexUiState.radialGeoSearch &&
-            indexUiState.radialGeoSearch.lat
-          ) {
-            ret.near = indexUiState.radialGeoSearch;
-          }
-
-          ret.q = indexUiState.query;
-
```
sherakama, line 116: Thanks for cleaning this up.

ericbakenhus:
@sherakama 

> Date options are looking really small IMHO

Are you running it locally? MUI's theming doesn't like HMR and loses styles. It looks like this without HMR:

Screenshot 2024-07-30 at 7 30 05 PM

> Clear filters is hidden under options

The search bar "clear search" or something else? I get the following in FF, Safari, and Chrome:

Screenshot 2024-07-30 at 7 33 23 PM


> Console error about switching to controlled vs uncontrolled on date filter

Looks like the custom toggle in the desktop date filter. I'll look at it separately.

>  Location mobile city search isn't working.

It's never worked locally (I assumed that was intentional?). Works fine in a deploy preview.

> Missing `refine` error

It's the same code as the desktop filter. I'll have to track it down.

> Location city and sub-filter options should have their drop-down changed to an inline element instead of an absolute element as it overflows

Yeah, I was trying to avoid spending time on it. 😆  I'll take care of it now.

> Sub-facets 'view results' button closes the first panel but not the second panel as well

I must have passed the wrong function to the third-level location drawers. Should be an easy fix.

sherakama:
> Are you running it locally? MUI's theming doesn't like HMR and loses styles. It looks like this without HMR:

yeah, let me re-run in preview


> Looks like the custom toggle in the desktop date filter. I'll look at it separately.

Yeah, not a blocker. 

> It's never worked locally (I assumed that was intentional?). Works fine in a deploy preview.

Huh, I'm pretty sure it was running locally for me at least at one point. I'll take a peek too.

> Yeah, I was trying to avoid spending time on it. 😆 I'll take care of it now.

I was going to wait to see what mobile looked like but my $.02 is that you should be able to disablePortal on mobile and set the `su-relative` in mobile. That hopefully should do the trick.

ericbakenhus:
@sherakama 

> Sub-facets 'view results' button closes the first panel but not the second panel as well

This one should be fixed now. Thanks! 👍

ericbakenhus:
@sherakama 

> Missing refine error

Does Algolia return undefined for `refine` when `canRefine` is false? Could that be what's happening here and we just need to add a `canRefine` check? I've never actually tested that...

ericbakenhus:
@sherakama City search works fine in the events discovery deploy preview but is borked here. I can't see any obvious changes that would cause that.

sherakama:
> I went ahead and tossed in a quick check to make sure it's defined before using it. It'll trigger a current location error if not, but that seems better than an unhanded exception.

I'm guessing it has something to do with scope/closures and the primary state. Troubleshooting some.

sherakama:
@ericbakenhus 

The TLDR; on the location bug was that multiple primary widgets were being created and that mixed up the information that was coming out of the hook.

https://github.com/SU-SWS/saa_alumni/pull/904/commits/5b987dd1748448963ab01e097d2c18cf0930e6cd has the checks to ensure only one primary is created and to ensure the refine methods are created as the render state was populated before the helper functions were created and added to the connector state.

src/components/events-discovery/components/Filters/MobileFilterFooter.jsx
ericbakenhus: Added. 👍

src/components/events-discovery/components/LocationFilter/MobileLocationFilter.jsx
ericbakenhus: Fixed. 👍

ericbakenhus:
@sherakama 

> https://github.com/SU-SWS/saa_alumni/commit/5b987dd1748448963ab01e097d2c18cf0930e6cd has the checks to ensure only one primary is created and to ensure the refine methods are created as the render state was populated before the helper functions were created and added to the connector state.

Thanks!

> Console error about switching to controlled vs uncontrolled on date filter

It's only happening locally. I've confirmed that the value we're passing to the datepicker is always null or a Luxon object so it shouldn't be switching based on our code. I think we're running into some MUI/double render in dev weirdness.

src/components/events-discovery/eventsDiscovery.jsx
```diff
@@ -19,15 +20,60 @@ import { FacetProvider } from './components/Facets/FacetCtx';
...
+    if (dateRefinements.length === 1) {
+      const refinement = dateRefinements[0];
+      return `(startTimestamp ${refinement.operator} ${refinement.value} OR endTimestamp ${refinement.operator} ${refinement.value}) AND (${baseFilter})`;
+    }
+
+    const startRefinement = dateRefinements.find((r) => r.operator === '>=');
+    const endRefinement = dateRefinements.find((r) => r.operator === '<=');
+
+    return `((startTimestamp:${startRefinement.value} TO ${endRefinement.value}) OR (endTimestamp:${startRefinement.value} TO ${endRefinement.value})) AND (${baseFilter})`;
+  }, [dateRefinements]);
```
ericbakenhus, line 75: @sherakama I tossed this bit of work in here as well. We have a request to consider `endTimestamp` as well when filtering based on date. I couldn't figure out how to properly set up a disjunctive filter across two facets without dropping down to the filter prop. Here's the code I'm using to generate the correct filter based on the `startTimestamp` refinement (I didn't want to blast away the existing date filter work).

src/components/events-discovery/eventsDiscovery.jsx
```diff
@@ -19,15 +20,60 @@ import { FacetProvider } from './components/Facets/FacetCtx';
...
+          ...(!!numericFilters?.length && numericFilters),
+        },
+      };
+    });
+
+    return algoliaClient.search(processedRequests);
+  },
+};
+
+const indexName = process.env.GATSBY_EVENT_ALGOLIA_INDEX;
```
ericbakenhus, line 49: @sherakama And here I'm intercepting the requests made to Algolia to remove any `startTimestamp` numeric filters since they're being handled in the filter prop.

ericbakenhus: Also swapped the hardcoded indexName with an env var.

ericbakenhus:
@sherakama 

> It's never worked locally (I assumed that was intentional?). Works fine in a deploy preview.

Ah, that was my bad. I didn't pull the latest env vars. It's working locally for me now as well. 👍

ericbakenhus:
@sherakama I think we've addressed (in some way) all the things found. Let me know if I'm overlooking anything. I'm going to go ahead and merge this to get it in front of the client for QA.

src/components/events-discovery/eventsDiscovery.jsx
sherakama: > how to properly set up a disjunctive filter across two facets

I don't think you can do that with the out-of-the-box instantsearch components and would likely need to handle the conditions in the middleware or in a custom connector.

src/components/events-discovery/eventsDiscovery.jsx
sherakama: Thanks for swapping out the var.

sherakama:
"Soft launch" === Let's get bug reports and fix them before we tell a lot of people about it.

src/components/events-discovery/eventsDiscovery.jsx
ericbakenhus: Before landing here, I looked into middleware but it seemed to only focus on manipulating the UI state. I also looked into a custom connector and, after holding in the vomit, didn't see an easy way to handle this. I found someone asking about a similar use case and the only suggestion from Algolia support was "toss it in the config filter." So, here we are. 😅 

A good candidate for seeing if we can upgrade from Good Enough™ to Not Terrible™.

PR #913, opened 2024-08-01
Nojira | @ericbakenhus | Adjust old story handling

ericbakenhus:
# READY FOR REVIEW

# Summary
- Adjusting some old event handling and fixing an id

# Review By (Date)
- Whenever

# Criticality
- High

# Review Tasks

1. Look at code

### DIFF
--- a/netlify/functions/event-import-background.mts
+++ b/netlify/functions/event-import-background.mts
@@ -228,8 +228,42 @@ export default async (req: Request) => {
       console.log('>>> Processing: ', id);
       try {
         if (google && storyblok) {
+          const isGoogleOld = luxonDate(
+            google.content.endOverride 
+            || google.content.end 
+            || google.content.startOverride 
+            || google.content.start
+          ) < OldCutoff;
+
+          const isSbOld = luxonDate(
+            storyblok.content.endOverride 
+            || storyblok.content.end 
+            || storyblok.content.startOverride 
+            || storyblok.content.start
+          ) < OldCutoff;
+
+          const isOld = isGoogleOld && isSbOld;
+
           console.log('Exists in Google and Storyblok...');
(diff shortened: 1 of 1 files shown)


netlify/functions/event-import-background.mts
```diff
@@ -228,8 +228,42 @@ export default async (req: Request) => {
...
+                  parent_id: eventArchiveFolderId,
+                },
+              });
+              await delay();
+            }
+            console.log('Done!');
+            console.log('Processing complete: ', id);
+            continue;
+          }
+
```
ericbakenhus, line 266: Old events are being cleared out of the Google sheets. We'll ask about it, but we should handle that case.

netlify/functions/event-import-background.mts
```diff
@@ -289,27 +325,26 @@ export default async (req: Request) => {
...
-          }
-
-          if (isOld) {
-            console.log('Story is old. Moving...');
-            if (run) {
               await storyblokManagement.put(`/spaces/${spaceId}/stories/${storyblok.id}`, {
                 story: { 
                   ...storyblok,
-                  full_slug: `events/sync-archive/${storyblok.slug}`,
+                  parent_id: eventArchiveFolderId,
```
ericbakenhus, line 336: Fixing this. Management API needs parent id.

netlify/functions/event-import-background.mts
sherakama: Huh, I definitely remember that being an expectation/assumption. Thanks for following up on it.

sherakama (approved):

ericbakenhus:
Really need some refactoring here. This is some wet ass code.

sherakama:
😆 I think that is the first time I have seen the words 'code' 'ass' and 'wet' used in that combination.

PR #923, opened 2024-08-08
DS-800 | @ericbakenhus | No context var for scheduled functions

ericbakenhus:
# READY FOR REVIEW

# Summary
- Remove context check in scheduled function

# Review By (Date)
- Retro

# Criticality
- High

# Review Tasks

1. Look at code

- DS-800

### DIFF
--- a/netlify/functions/trigger-event-import.mts
+++ b/netlify/functions/trigger-event-import.mts
@@ -8,11 +8,6 @@ export default async (req: Request) => {
     const secret = process.env.EVENT_IMPORT_SECRET ?? '';
     console.log('=== START Trigger Event Import ===');
 
-    // Make sure we only run this in production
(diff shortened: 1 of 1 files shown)


netlify/functions/trigger-event-import.mts
```diff
@@ -8,11 +8,6 @@ export default async (req: Request) => {
     const secret = process.env.EVENT_IMPORT_SECRET ?? '';
     console.log('=== START Trigger Event Import ===');
 
-    // Make sure we only run this in production
-    if (process.env.CONTEXT !== 'production') {
-      throw new Error('Not in prod');
-    }
```
ericbakenhus, line 14: This isn't populated. Kinda makes sense since scheduled functions only run in prod.

PR #943, opened 2024-09-04
Nojira | @ericbakenhus | Fix maps key

ericbakenhus:
# READY FOR REVIEW

# Summary
- Fix google key
- Remove region comparison check

# Review By (Date)
- ASAP/Retro

# Criticality
- High

# Review Tasks

1. Look at code

### DIFF
--- a/netlify/functions/event-import-background.mts
+++ b/netlify/functions/event-import-background.mts
@@ -274,18 +274,20 @@ export default async (req: Request) => {
 
           console.log('Changes detected. Syncing changes to Storyblok...');
 
+          console.log({ id, isMissingRegion, google, storyblok });
+
           if (run) {
-            console.log('Generating region...');
-            const combinedStory = await setStoryRegion(combineStories(google, storyblok), googleMapsKey);
-            console.log(combinedStory.content.region ? 'Generating region done!' : 'Generating region failed!');
-            await storyblokManagement.put(`/spaces/${spaceId}/stories/${storyblok.id}`, {
-              story: {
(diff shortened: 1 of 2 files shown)


netlify/functions/event-import-background.mts
```diff
@@ -276,7 +276,7 @@ export default async (req: Request) => {
 
           if (run) {
             console.log('Generating region...');
-            const combinedStory = await setStoryRegion(combineStories(google, storyblok));
+            const combinedStory = await setStoryRegion(combineStories(google, storyblok), googleMapsKey);
```
ericbakenhus, line 279: I'm going to blame the general concept of JS here (instead of me forgetting to add some JSDoc).

src/utilities/synchronizedEvents.js
```diff
@@ -359,7 +359,6 @@ export const compareStoryContent = (a, b) => {
     a.state !== b.state ||
     a.country !== b.country ||
     a.address !== b.address ||
-    a.region !== b.region ||
```
ericbakenhus, line 362: Region is now generated during import and shouldn't be included in the comparison check.

netlify/functions/event-import-background.mts
sherakama: How's that Nextjs rebuild with TypeScript going? 

Man, the more I work in this repo, the more I miss TS.

sherakama (approved):
Don't let the right hand know what the left hand is doing.

PR #945, opened 2024-09-04
HOM-2 small interior banners for homesite

shc314:
# READY FOR REVIEW

# Summary
- This implements the HeaderMinimal banner component, which is based on the one with the same name from the OOD Interior Page type.

# Review By (Date)
- When possible

# Review Tasks

1. Check out this branch
2. Place the Header Minimal block into the Above Content section of a basic page.
3. Verify that the background color can be set, a title can be added, and the spacing below the header can be adjusted.

## General
- This is one of the four header styles based on the ones on the OOD Giving site. I want to make sure that I'm approaching it correctly before proceeding with the remaining three.

- JIRA ticket: [HOM-2](https://internal-saa.atlassian.net/browse/HOM-2?atlOrigin=eyJpIjoiN2RkZmRiMjE1NWI5NDQ4YjhkODU0MTBkZGI2MTA5YmYiLCJwIjoiaiJ9)
- @willyumfrancis

### DIFF
--- a/src/components/partials/centeredContainer.js
+++ b/src/components/partials/centeredContainer.js
@@ -1,18 +1,25 @@
 import React from 'react';
 
-const CenteredContainer = (props) => {
-  const Element = props.element ?? 'div';
+const CenteredContainer = ({
+  element,
+  centeredDisabled,
+  flex,
+  srOnly,
+  classes,
+  children,
+}) => {
+  const Element = element ?? 'div';
 
   return (
     <Element
       className={`
-       ${props.centered_disabled ? '' : 'centered-container'}
-       ${props.flex ? 'flex-container' : ''}
-       ${props.srOnly ? 'su-sr-only-element' : ''}
-       ${props.classes ?? ''}
+       ${centeredDisabled ? '' : 'su-cc'}
+       ${flex ? 'flex' : ''}
+       ${srOnly ? 'su-sr-only-element' : ''}
+       ${classes ?? ''}
   `}
     >
-      {props.children}
+      {children}
     </Element>
   );
 };
--- a/src/components/partials/centeredContainer.js
+++ b/src/components/partials/centeredContainer.js
@@ -1,0 +1,20 @@
+import React from 'react';
+
+const CenteredContainer = (props) => {
+  const Element = props.element ?? 'div';
+
+  return (
+    <Element
+      className={`
+       ${props.centered_disabled ? '' : 'centered-container'}
+       ${props.flex ? 'flex-container' : ''}
+       ${props.srOnly ? 'su-sr-only-element' : ''}
+       ${props.classes ?? ''}
+  `}
+    >
+      {props.children}
+    </Element>
+  );
+};
+
+export default CenteredContainer;
--- a/src/components/partials/heading.js
+++ b/src/components/partials/heading.js
@@ -1,0 +1,22 @@
+import React from 'react';
+
+const Heading = (props) => {
+  const HeadingTag = props.level || props.defaultLevel || 'h3';
+
+  return (
+    <HeadingTag
+      className={`
+                ${props.classes ? props.classes : ''}
+                ${props.serif ? 'su-serif' : ''}
+                ${props.weight ? `su-${props.weight}` : ''}
+                ${props.color ? `su-text-${props.color}` : ''}
+                ${props.align ? `su-text-align-${props.align}` : ''}
+                ${props.external ? 'su-link--external' : ''}
+    `}
+    >
+      {props.children}
+    </HeadingTag>
+  );
+};
+
+export default Heading;
--- a/src/components/partials/heading.js
+++ b/src/components/partials/heading.js
@@ -1,20 +1,30 @@
 import React from 'react';
 
-const Heading = (props) => {
-  const HeadingTag = props.level || props.defaultLevel || 'h3';
+const Heading = ({
+  level,
+  defaultLevel,
+  classes,
+  serif,
+  weight,
+  color,
+  align,
+  external,
+  children,
+}) => {
+  const HeadingTag = level || defaultLevel || 'h3';
 
   return (
     <HeadingTag
       className={`
-                ${props.classes ? props.classes : ''}
-                ${props.serif ? 'su-serif' : ''}
-                ${props.weight ? `su-${props.weight}` : ''}
-                ${props.color ? `su-text-${props.color}` : ''}
-                ${props.align ? `su-text-align-${props.align}` : ''}
-                ${props.external ? 'su-link--external' : ''}
+                ${classes || ''}
+                ${serif ? 'su-font-serif' : ''}
+                ${weight ? `su-font-${weight}` : ''}
+                ${color ? `su-text-${color}` : ''}
+                ${align ? `su-text-${align}` : ''}
+                ${external ? 'su-link--external' : ''}
     `}
     >
-      {props.children}
+      {children}
     </HeadingTag>
   );
 };
--- a/src/hooks/useWindowSize.js
+++ b/src/hooks/useWindowSize.js
@@ -2,7 +2,7 @@ import { useState, useEffect } from 'react';
 
 /**
  * @see: https://usehooks.com/
- **/
+ * */
 
 function UseWindowSize() {
   // Initialize state with undefined width/height so server and client renders match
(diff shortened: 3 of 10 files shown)


yvonnetangsu:
Hi @shc314 👋🏼  Just curious - why is there such a large change in the `package-lock.json` file without a `package.json` update? Did you recently update your packages? Personally I usually go with a `npm ci` instead of `npm install` so I'm using the existing `package-lock.json` versions just to be safe - any concerns here @sherakama?

shc314:
Hi @yvonnetangsu , thanks for checking. I was having a bunch of errors so @sherakama advised me to remove package-lock.json and install the packages again. Perhaps he can explain it better.

src/components/partials/centeredContainer.js
```diff
@@ -0,0 +1,27 @@
...
+       ${srOnly ? 'su-sr-only-element' : ''}
+       ${classes ?? ''}
+  `}
+    >
+      {children}
+    </Element>
+  );
+};
+
+export default CenteredContainer;
```
yvonnetangsu, line 27: @shc314 Is there a reason not to use the existing `Container` component instead of creating this new one? It seems like `Container` could do the job of this new component.

https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/layout/Container.jsx

src/components/partials/heading.js
```diff
@@ -0,0 +1,32 @@
...
+                ${align ? `su-text-${align}` : ''}
+                ${external ? 'su-link--external' : ''}
+    `}
+    >
+      {children}
+    </HeadingTag>
+  );
+};
+
+export default Heading;
```
yvonnetangsu, line 32: Similarly, I don't think you need to create this new heading component since we have an existing one:
https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/simple/Heading.jsx

src/components/partials/centeredContainer.js
shc314: Thanks for pointing that out. I didn't know the Container component existed. I'll try to get it to work.

src/components/partials/heading.js
shc314: Likewise, I didn't know it existed. Thanks for bringing it to my attention.

yvonnetangsu:
@sherakama What do you think about the `package-lock.json` change in this PR? Could @shc314 just do a `npm ci` (with the fontawesome token in front) instead (so we don't have to commit the 30,000 lines of code change)? 😄 @shc314 mentioned a problem with her local install earlier.

sherakama:
> @sherakama What do you think about the `package-lock.json` change in this PR? Could @shc314 just do a `npm ci` (with the fontawesome token in front) instead (so we don't have to commit the 30,000 lines of code change)? 😄 @shc314 mentioned a problem with her local install earlier.

Yeah, normally I'd be wary of the `package-lock.json` change as well but I did work with @shc314 through the environment issues and this was the resulting set of versions. Let me pull this down real quick, validate versions and push back up a GTG version.

sherakama:
Ok, down to a _reasonable_ `1,908` line change.

yvonnetangsu:
> Ok, down to a _reasonable_ `1,908` line change.

Thanks for looking at this @sherakama. That is better than 30K lines indeed 😆

sherakama:
@shc314, from what I can tell it looks like you have a few revisions to this PR. Let's us know if you have any questions.

shc314:
@yvonnetangsu @sherakama It's ready for review again. Thank you.

src/hooks/useWindowSize.js
sherakama, line 1: @shc314 `useMediaQuery` and `useDisplay` hooks already exist in this repo. Can you tell me more about the need for this hook?

yvonnetangsu:
@shc314 FYI for future PRs, I believe you can publish any pages on the DEV Storyblok space so a preview link can be shared more easily - is there concerns with doing that? Just thinking that it would be easier for reviews. cc @sherakama

sherakama:
> @shc314 FYI for future PRs, I believe you can publish any pages on the DEV Storyblok space so a preview link can be shared more easily - is there concerns with doing that? Just thinking that it would be easier for reviews. cc @sherakama

I don't have any issues with @shc314 creating and publishing dev content for reviews. In fact, that is highly encouraged.

yvonnetangsu:
> > @shc314 FYI for future PRs, I believe you can publish any pages on the DEV Storyblok space so a preview link can be shared more easily - is there concerns with doing that? Just thinking that it would be easier for reviews. cc @sherakama
> 
> I don't have any issues with @shc314 creating and publishing dev content for reviews. In fact, that is highly encouraged.

In that case, it would be awesome if you could publish preview pages on the DEV space and include the links in the "Steps to test" in the PR 😄 🙏🏼  This would likely speed up the reviewing process so you can get things into production sooner.

src/hooks/useWindowSize.js
shc314: I didn't know they existed. I'll use them instead. Thanks for bringing this to my awareness.

shc314:
> > > @shc314 FYI for future PRs, I believe you can publish any pages on the DEV Storyblok space so a preview link can be shared more easily - is there concerns with doing that? Just thinking that it would be easier for reviews. cc @sherakama
> > 
> > 
> > I don't have any issues with @shc314 creating and publishing dev content for reviews. In fact, that is highly encouraged.
> 
> In that case, it would be awesome if you could publish preview pages on the DEV space and include the links in the "Steps to test" in the PR 😄 🙏🏼 This would likely speed up the reviewing process so you can get things into production sooner.

@yvonnetangsu Will do. I was trying to share the link just now. I know I shouldn't include the access key when posting the link, but I can't get it to work without it. https://deploy-preview-945--stanford-alumni.netlify.app/interior-banners-test gets a 404 error

shc314:
> @shc314 You wouldn't need to include the access key if the page is published. The reason that the link was showing a 404 was because it was published after any code commit from this PR was pushed up. I triggered a rebuild for you on Netlify so it's working now. Next time if you publish before you push up a commit you'll be able to see the page 😄

@yvonnetangsu Got it. Thank you!

yvonnetangsu (changes requested):
@shc314 👋🏼 Looking at this more closely, I believe this interior banner could be built out a bit differently. Instead of creating it as a Storyblok component that can be nested inside the Above Content region, it could be rendered as part of the Basic Page component. For example, we currently already have a setting to show the generic H1 on a white background with no image like this - this is with the `"Only check this if a Hero banner is present. Make page title screen reader only (not visible)"` box unchecked:

Screenshot 2024-10-01 at 3 06 44 PM

The code that renders this generic page header is here in this file:
https://github.com/SU-SWS/saa_alumni/blob/dev/src/components/page-types/basicPage.js#L76

If you add a field simply to the Basic Page for the user to select the hero background color, then you could just render this new interior banner using that info, and with some changes to the Basic Page file. This way you won't end up with 2 h1's on the page, and the authors would not have to add it as a nested component. It will be using the "Page title (h1)" field from the Basic page + your new background color option field to render it.

cc @sherakama

shc314:
@yvonnetangsu Thanks for that suggestion. The Art Director basically wants to replicate the HeaderMinimal, HeaderNoImage, HeaderWithImage, and HeaderFullWidthImage that are on the OOD Giving site. Can the other three headers be built similarly?

yvonnetangsu:
> @yvonnetangsu Thanks for that suggestion. The Art Director basically wants to replicate the HeaderMinimal, HeaderNoImage, HeaderWithImage, and HeaderFullWidthImage that are on the OOD Giving site. Can the other three headers be built similarly?

@shc314 Yes that can be built similarly 😄 In that case, I'd probably add a single option field that let the user pick which hero/header style they want. There's already a hero image field in Storyblok, so you can make sure of that to build out the variants with image. For the other style, you could add a background color field as discussed before. With data you collect from these new fields + the existing hero image field, you can then add conditionals based on what hero style they pick, and render the appropriate one in the Basic Page component.

shc314:
@yvonnetangsu My changes are ready for review. So far I've only implemented Header Minimal and Header No Image, so even though "has image" and "full width image" are options in the Header Style dropdown menu, they won't do anything at this point.

For headerNoImage, I wasn't able to get the size of the title and the space between it and the intro text box to match exactly what's on the Giving site. Do you have advice?

PR #973, opened 2024-09-25
DS-987 | @ericbakenhus | Filter search suggestions based on full results filter

ericbakenhus:
# READY FOR REVIEW

# Summary
- Adds same base filter used by the results to the search bar
    - This will prevent expired events and events outside the current date filter range from being suggested
- (Tossed in) Fix deployment function so it's able to get unpublished stories in order to remove them from Algolia

# Review By (Date)
- Now/retro

# Criticality
- High

# Review Tasks

1. Check out this branch
3. Search for Hong Kong
4. Verify there are no suggestions

- DS-987

### DIFF
--- a/netlify/functions/handle-deploy.mts
+++ b/netlify/functions/handle-deploy.mts
@@ -47,7 +47,7 @@ export default async (req: Request) => {
     }
 
     const data: SBWebhookPayload = await JSON.parse(rawData);
-    console.log(data.text);
+    console.log(data);
     
     if (data.action !== 'published' && data.action !== 'unpublished') {
       // Trigger rebuild and stop
@@ -70,7 +70,9 @@ export default async (req: Request) => {
       oauthToken: process.env.STORYBLOK_MANAGEMENT_OAUTH_TOKEN,
     });
 
-    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`);
+    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`, {
+      version: data.action === 'published' ? 'published' : 'draft',
+    });
     const story = storyRes?.data?.story;
     const isFolder = story?.is_folder;
     const contentType = story?.content?.component;
@@ -109,7 +111,7 @@ export default async (req: Request) => {
       storiesToProcess = await storyblokContent.getAll('cdn/stories', { 
         starts_with: 'events/sync/', 
         content_type: 'synchronizedEvent', 
-        version: data.action === 'published' ? 'draft' : 'published' 
+        version: data.action === 'published' ? 'published' : 'draft' 
       }) ?? [];
 
       console.log('Event folder (un)publish event detected. Deploying...');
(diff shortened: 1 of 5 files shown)


src/components/events-discovery/eventsDiscovery.jsx
```diff
@@ -88,12 +88,17 @@ const EventDiscoveryContent = () => {
...
+              searchClient={searchClient}
+              indexName={indexName}
+              filters={filters}
+            />
           </div>
           <div className="su-block lg:su-hidden su-w-full">
             <MobileSearchBar
               searchClient={searchClient}
               indexName={indexName}
+              filters={filters}
```
ericbakenhus, line 101: Passing in the same filter used by the search results. No more expired events or events outside the current date range selection.

Might be good to refactor this in a future PR so all the current filters are taken into account.

src/components/events-discovery/components/Facets/FacetList.jsx
```diff
@@ -71,7 +71,7 @@ export const FacetList = ({ attribute, subfilter = false }) => {
     >
       {canRefine ? (
         <>
-          <DynamicWidgets>
+          <DynamicWidgets maxValuesPerFacet={1000}>
```
sherakama, line 74: This is an increase from the default, correct? We're not expecting over 1000.

netlify/functions/handle-deploy.mts
```diff
@@ -70,7 +70,9 @@ export default async (req: Request) => {
       oauthToken: process.env.STORYBLOK_MANAGEMENT_OAUTH_TOKEN,
     });
 
-    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`);
+    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`, {
+      version: data.action === 'published' ? 'published' : 'draft',
+    });
```
ericbakenhus, line 75: Can't grab the event if it's just been unpublished without asking for the draft version.

netlify/functions/handle-deploy.mts
```diff
@@ -109,7 +111,7 @@ export default async (req: Request) => {
       storiesToProcess = await storyblokContent.getAll('cdn/stories', { 
         starts_with: 'events/sync/', 
         content_type: 'synchronizedEvent', 
-        version: data.action === 'published' ? 'draft' : 'published' 
+        version: data.action === 'published' ? 'published' : 'draft'
```
ericbakenhus, line 114: I'm pretty sure this was backwards...

netlify/functions/handle-deploy.mts
```diff
@@ -47,7 +47,7 @@ export default async (req: Request) => {
     }
 
     const data: SBWebhookPayload = await JSON.parse(rawData);
-    console.log(data.text);
+    console.log(data);
```
ericbakenhus, line 50: Just log the entire SB payload.

src/components/events-discovery/components/Facets/FacetList.jsx
ericbakenhus: I can't imagine there would ever be more than 1000 subjects...

We're dynamically changing the limit in the child widget to handle the more/less button. The warning suggested that since the default value was below the limit, it might trigger two network request. So I set it to 1000 to match the larger limit value.

sherakama (approved):
Thanks

PR #975, opened 2024-09-25
Nojira | @ericbakenhus | Fix deploy

ericbakenhus:
# READY FOR REVIEW

# Summary
- Actually fix deploy issues
    - Revert bad options change
    - Move a filter that was preventing expired stories from being removed from Algolia

# Review By (Date)
- Soon/Retro

# Criticality
- High

# Review Tasks

1. Look at code?

### DIFF
--- a/netlify/functions/handle-deploy.mts
+++ b/netlify/functions/handle-deploy.mts
@@ -70,9 +70,8 @@ export default async (req: Request) => {
       oauthToken: process.env.STORYBLOK_MANAGEMENT_OAUTH_TOKEN,
     });
 
-    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`, {
-      version: data.action === 'published' ? 'published' : 'draft',
-    });
+    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`);
+    console.log(storyRes);
     const story = storyRes?.data?.story;
     const isFolder = story?.is_folder;
     const contentType = story?.content?.component;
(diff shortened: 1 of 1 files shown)


netlify/functions/handle-deploy.mts
```diff
@@ -70,9 +70,7 @@ export default async (req: Request) => {
       oauthToken: process.env.STORYBLOK_MANAGEMENT_OAUTH_TOKEN,
     });
 
-    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`, {
-      version: data.action === 'published' ? 'published' : 'draft',
-    });
+    const storyRes = await storyblokManagement.get(`/spaces/${data.space_id}/stories/${data.story_id}`);
```
ericbakenhus, line 73: Forgot the management API doesn't support those options and just returns everything.

netlify/functions/handle-deploy.mts
```diff
@@ -111,18 +109,18 @@ export default async (req: Request) => {
...
+
       console.log('Event folder (un)publish event detected. Deploying...');
       if (run) {
         await fetch(deployUrl, { method: 'POST' });
       }
       console.log('Deploy triggered');
     }
 
-    storiesToProcess = storiesToProcess.filter((story) => story.full_slug.startsWith('events/sync/'));
-
```
ericbakenhus, line 125: Moving this inside the event folder conditional block. Only in that specific case do we want to be upserting/deleting stories specifically in the event/sync folder.

ericbakenhus:
Need to figure out better ways to test these functions...

sherakama:
> Need to figure out better ways to test these functions...

We can chat more about this in our Monday developer chat

PR #990, opened 2024-10-10
DS-998 | @ericbakenhus | Handle time zone changes that cross day boundaries

ericbakenhus:
# READY FOR REVIEW

# Summary
- Handle edge case where changing an event's displayed time zone to the user's local time zone crosses a date boundary
    - Switch the check for single/multi day events to look only at the event time zone and not the selected time zone
    - Display a +/- 1 day flag when a time crosses into a different day

# Review By (Date)
- Soon

# Criticality
- Medium

# Review Tasks

1. Check out this branch
3. Change your system's time zone as needed
4. Swap single day event time zones to cause start or end time to cross into a different day
5. See the new +/- 1 day flag

For example:
Screenshot 2024-10-10 at 2 16 49 PM
Screenshot 2024-10-10 at 2 16 55 PM

- DS-998

### DIFF
--- a/src/components/events-discovery/components/Event/EventContent.jsx
+++ b/src/components/events-discovery/components/Event/EventContent.jsx
@@ -1,6 +1,6 @@
 import React, { useId, useMemo, useState } from 'react';
 // eslint-disable-next-line no-unused-vars
-import { DateTime } from 'luxon';
+import { DateTime, Interval, Duration } from 'luxon';
 import { MenuItem, Select } from '@mui/material';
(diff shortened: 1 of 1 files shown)


ericbakenhus:
@sherakama I'm double checking on the copy, but everything else should be ready. 👍

sherakama (approved):

sherakama:
Screenshot 2024-10-10 at 16 57 35

Looks good from Japan.

